PR #16, opened 2015-06-19
Specify what to do when enter is pressed in various input fields

jtormey:
Contains fix for WEBHD-199


Sjors:
Can you make this pull request to dev?

PR #22, opened 2015-06-24
Hide settings

jtormey:
Hide certain settings that are awaiting a long term solution


Sjors:
Both issues are solved now, so this patch is no longer needed. Thanks though. :-)

PR #57, opened 2015-07-24
Prevent selection from happening when clicking toggle tx info button

jtormey:
Double clicking the "toggle" button on a transaction item will cause some annoying highlight. If this is the right way to fix, please merge


keonne:
outline:0; is what you are looking for I think Justin.

PR #82, opened 2015-09-07
Typography

abhisharma2:
No screenshots since all I did here was change the type treatment across all the views of the app. Overkill to see the diff. Youmna will be running a final audit to make sure we're normalized on typography going forward


Sjors:
Can you rebase it on the latest master and ping me? I'll merge it then.

PR #90, opened 2015-09-08
fix(UI): receive modal input widths auto sizing issues

abhisharma2:
Fixes the input field from overflowing past it's parent element in Firefox. Left aligned text while I was in here, too
screen shot 2015-09-08 at 3 01 22 pm


Sjors:
There was a merge conflict:
merge
Let me know if I fixed it incorrectly.

PR #98, opened 2015-09-09
feat(HTML5): URL's without #

Sjors:
Uses `$locationProvider.html5Mode(true)` as well as the `<base>` tag. The
base tag is modified by Grunt for distribution.

Requires server-side changes; these are implemented in dev mode. See [this](https://github.com/angular-ui/ui-router/wiki/Frequently-Asked-Questions#how-to-configure-your-server-to-work-with-html5mode) for hints on how to configure the server so that refreshing a URL works.


Sjors:
Rebased. One thing that's a bit annoying is that we have enumerate all AngularJS routes on the server. That way index.html is rendered for e.g. `blockchain.info/wallet-beta/1/transactions`, but not for a 404 like `blockchain.info/wallet-beta/1/transac` or for `blockchain.info/wallet-beta/assets/*`. This means making server-side changes whenever we change a route, or we have to do something smart.

Sjors:
I simplified things a bit, although the fact that I changed 40 files suggests otherwise :-)

No more need to maintain a list of AngularJS routes on the server. All images, css, etc are now in `/assets`. So the server just needs to consider three kinds of paths:

* `/wallet-beta/assets/{1}` -> `/dist/{1}` (return 404 if nothing is found please)
* `/wallet-beta/*` -> `index.html` (AngularJS will do the right thing is it can't find a route)
* `/wallet-beta` -> redirect to `/wallet-beta/` (we already do that)

Sjors:
Rebased for semistandard.

Sjors:
Rebased **again** for #399 and some other changes that affect asset URL's.

PR #102, opened 2015-09-10
modal_notification_ctrl converted to es6 

karinashenaz:


assets/js/controllers/modal_notification_ctrl.js
```diff
@@ -0,0 +1,4 @@
+walletApp.controller("ModalNotificationCtrl", ($scope, Wallet, $modalInstance, notification) => {
+  $scope.notification = notification;
+  $scope.ok = () => $modalInstance.close(notification);
```
Sjors: I prefer to have this function enclosed with brackets and on it's own line (without a return statement).

karinashenaz: done

PR #106, opened 2015-09-10
security_center_ctrl converted to es6

karinashenaz:


assets/js/controllers/security_center_ctrl.js
```diff
@@ -0,0 +1,120 @@
...
+      }
+    };
+  };
+
+  $scope.transfer = address => {
+    $modal.open({
+      templateUrl: "partials/send.jade",
+      controller: "SendCtrl",
+      resolve: {
+        paymentRequest: () => {
```
Sjors: If you replace { with ( then you can rid of the return statement.

PR #110, opened 2015-09-10
Add mock balance history chart

GregDracoulis:
Depends on https://github.com/blockchain/My-Wallet-V3/pull/34

I'd note that in the future, it's probably better to return a sorted daily average (or ending) balance from the API itself, rather than requiring the frontend to sort and filter unix timestamps. `angular-charts` seems to get mad if you try giving it empty data points to actually space out dates, so it'd be nice to not have to fill in the blanks in the frontend.

Ideally, you'd want to specify `start` and `end` query params with unambiguous dateints to specify a date range, so you could show the balance over "Last 7 Days", "Last 30 Days", "August", etc.

TODO: Space dates out evenly by creating points for missing dates, internationalize month abbreviations.

### DIFF
--- a/app/partials/home.jade
+++ b/app/partials/home.jade
@@ -20,3 +20,13 @@
               ac-config="pieChartConfig"
               ng-show="accounts().length > 1 && getTotal() > 0"
             )
+      .section
+        .charts.pos-rel
+          h5.aaa.em-400.mbl(translate="BALANCE_OVER_TIME")
+          .widget.flex-center.flex-justify.pos-rel(ng-class="{ 'empty' : !balanceHistory().length}")
+            #balance-history-chart.chart(
+              ac-chart="'line'"
+              ac-data="lineChartData"
+              ac-config="lineChartConfig"
+              ng-show="balanceHistory().length > 1"
+            )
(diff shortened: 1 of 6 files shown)


GregDracoulis:
Now converts BTC to the fiat currency defined in settings. If that isn't available or doesn't have a conversion, defaults to displaying in BTC.

Also depends on https://github.com/blockchain/My-Wallet-V3/pull/35 now.

Sjors:
Hi @GregDracoulis, thanks for these pull requests!

I think the blockchain/My-Wallet-V3#34 part of your code works, but the frontend is throwing an error message and doesn't display the chart:

schermafbeelding 2015-09-11 om 12 10 05

The array you see in the console is Wallet.balanceHistory() at line 100. Let me know if you need more info for debugging.

GregDracoulis:
@Sjors just made a commit to quiet those console errors, and a couple others to fix the real issue. The errors were actually just a red herring-`angular-charts` would get mad initially if you supplied a `series` in its config data. Looks like it would try to loop through the empty data and throw an error if there were no points in it. Fixed by just supplying an empty object for the chart data initially.

The actual issue with the chart not displaying is fixed in https://github.com/blockchain/My-Wallet-V3/pull/35. If the converted balance is over 999.99, the requests to `getFiatAtTime` will error. Since I was using `$q.all` this means the rendering of the whole chart would fail if any request fails. I added a fallback error handler that reprocesses the data points in this case and just skips the conversion, displaying the whole chart in BTC if there's an error.

Sjors:
Ah of course, I should have merged blockchain/My-Wallet-V3#35 on my local machine, my bad.

GregDracoulis:
No worries, thanks for pointing it out! Definitely good to get rid of those console errors & to add an extra layer of error handling to the chart in case one of the requests gets zapped :smile:

PR #120, opened 2015-09-17
feat(Xpub): display warning to user before revealing Xpub

abhisharma2:
screen shot 2015-09-17 at 12 07 27 pm


kristovatlas:
ACK this PR. thanks for including a test with the commit

PR #126, opened 2015-09-21
Implement payment.js

jtormey:


abhisharma2:
Looks good - tested for the following scenarios:

Legacy -> External, Legacy -> Account,  Account -> External, Account -> Multiple External, Account -> Legacy

PR #142, opened 2015-09-29
fix(Send): press enter does the right thing

Sjors:
The destinations list is filtered as the user types. If the user copy-pastes an address that will be the only option left. That way if they hit enter, the correct address is selected (rather than falling back to the first account).

### DIFF
--- a/app/partials/send.jade
+++ b/app/partials/send.jade
@@ -58,7 +58,7 @@
               )
                 ui-select-match(placeholder="Paste, scan or select address") {{ $select.selected.label }}
                 ui-select-choices(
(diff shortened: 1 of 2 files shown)


jtormey:
Using `| filter: $select.search` in `ui-select-choices` causes a visual glitch where the dropdown flickers briefly as you type. Abhi and I fixed this by using the `filterDestinations` function rather than the angular filter. We probably want to pass `$select.search` into this function and handle the filtering there to prevent this glitch, unless you see another way.

Sjors:
I'll try passing it in.

PR #143, opened 2015-09-29
Only ask 2nd password once when importing BIP38

Sjors:
Wallet.addAddressOrPrivateKey() was called twice if a BIP38 encoded key was imported. As in the past I'm using a callback to prevent that. There might be a way to replace this callback hell with promises.


jtormey:
In general I'm not a huge fan of mixing promises and callbacks. Even though this callback has a very defined purpose, keeping track of these functions can be tricky and I'd like to try to avoid going back to what we were doing before if possible.

PR #165, opened 2015-10-06
Sign message for legacy address

Sjors:
`signMessage()` has not yet been refactored in My-Wallet-V3

schermafbeelding 2015-10-06 om 16 45 32


jtormey:
Moved to #381

PR #171, opened 2015-10-13
Removed jQuery and and replaced mobile number component

Sjors:
In settings:
settings

Country flags:
country flags

2FA modal:
2fa modal

Security center:
security center


jtormey:
Selecting countries from the flag dropdown messes with the page url. Probably because they are anchor tags with href="#". This needs to be handled by the dependency.

PR #176, opened 2015-10-21
Toggle advanced settings

Sjors:
Advanced hidden:
advanced hidden

Advanced shown:
advanced shown


keonne:
Few things:
- Not sure I like the text well underneath a settings option with a toggle button inside of it.
-How does one toggle advanced settings off?
- Is the orange color button brand approved?

PR #183, opened 2015-10-22
Shaky modal

abhisharma2:
1. Looks like something with ngAnimate went haywire (maybe with the bootstrap update? not sure really why). Need to add some hard and fast display none/blocks to make it lose the shake.

2. Changed up the grey in the Send modal to match with the rest of the greys ;)

3. Updated README to include the bower step

Sorry for including this all in one branch, but minor things.


jtormey:
Nice! I never would have suspected ngAnimate was the problem, you probably just saved me from hours of stressful debugging

PR #190, opened 2015-11-02
feat(recovery): add wallet recovery from login screen

jtormey:


jtormey:
A current downside of this feature is that it doesn't play nicely with the beta invite system. I think it is worth adding in now though, for QA/product team purposes and making sure the code doesn't go stale

PR #198, opened 2015-11-13
refactor(server): convert server.coffee to js, upgrade to express 4

jtormey:
This might also require changes to the dependency whitelist. Express 4 doesn't have built-in middleware, so I needed to install new middleware to replace what we had.


Sjors:
Is this safe to merge or should I hold off until you've rebased it?

jtormey:
Safe to merge. I had to add in the changes you made for the ip whitelist, it's worth looking that over in case I missed something (although I tested and the whitelist worked for me).

PR #209, opened 2015-12-01
Show when transaction is sent to/from watch-only address

Sjors:
watch only

Depends on blockchain/My-Wallet-V3#104


Sjors:
This should be good to go now. I also fixed the Dutch translation to longer refer to a wrist watch.

PR #224, opened 2015-12-08
fix(tasks): wrap dist min file in iife

jtormey:
Wraps our application in a closure, so as to not expose any functions to the global scope in production mode.


jtormey:
The functions this change hides are controller and service functions. The global "Blockchain" variable is still global.

PR #231, opened 2015-12-17
Signup validation

jtormey:
Fixes some validation messaging issues. Validation is now done the angular way, rather than in the controller. Also switches all `$cookieStore` (deprecated) references to `$cookies`.


jtormey:
The tests should all pass once the tests in `beta` are passing.

PR #236, opened 2016-01-10
Fix syntax error in recovery phase modal

plondon:


Sjors:
Thanks! Can you make this PR to the beta `branch` instead? That's where the action is at the moment.

PR #239, opened 2016-01-14
Single click copy

plondon:


assets/js/directives/single-click-select.js.coffee
```diff
@@ -21,6 +21,7 @@ angular.module('walletApp').directive('singleClickSelect', ($window) ->
 
         if scope.browserCanExecCommand
           $window.document.execCommand('copy')
+          scope.$digest()
```
jtormey: Calling $digest directly can throw an error if angular is already in a $digest loop. Generally, it's better practice to call the $safeApply function defined in app.js: `$scope.$root.$safeAppy()`. This will do the same thing, but only if necessary. [Article with more info](http://www.yearofmoo.com/2012/10/more-angularjs-magic-to-supercharge-your-webapp.html#apply-digest-and-phase).

Sjors:
Agree with @jtormey, looks good otherwise.

Sjors:
Looks good now, don't worry about the decreased test coverage. Can you do a git rebase on beta, so that Github knows how to merge it?

```sh
git checkout beta
git pull
git checkout single-click-copy
git rebase beta
```
If there are no conflicts:
```sh
git push origin single-click-copy --force
```

PR #242, opened 2016-01-14
Advanced send modal helper text

plondon:


Sjors:
Just noticed this after merge, but for next time the commit message needs a colon:
`fix(AdvancedSend) helper text visible...` should be:
`fix(AdvancedSend): helper text visible...`

Otherwise it will get ignored by Grunt when generating the changelog.

PR #246, opened 2016-01-15
Rocket gif alignment and aspect ratio

plondon:
Ensuring that the rocket gif does not stretch with the width of screen on resize


jtormey:
Nice

PR #253, opened 2016-01-20
Upgrade dialogue slider

plondon:
The best way to test this would be to add the following to somewhere that always gets called. Like login.controller

`const modalInstance = $uibModal.open({
    templateUrl: "partials/upgrade.jade",
    controller: "UpgradeCtrl",
    backdrop: "static",
    windowClass: "bc-modal"
  });`


Sjors:
I unmerged it, but Github doesn't see that. Can you open a new PR?

PR #261, opened 2016-01-22
Success notifications

plondon:
Clean up the success modal
screen shot 2016-01-22 at 12 36 06 pm

Successful send now mimics successful receive alert message
send success


Sjors:
Code looks good to me.

PR #265, opened 2016-01-22
China

Sjors:
It was not possible to switch to Chinese if the browser language was not Chinese. I believe I fixed it now so that:

1 - when the browser language is Chinese, the registration form guesses Chinese as the language and sets the correct setting
2 - if the browser language is English, the user can switch to Chinese

I only tried this with simplified chinese, not with traditional chinese.


Sjors:
Wrong branch.

PR #267, opened 2016-01-22
Accounts ui

plondon:

### DIFF
--- a/app/partials/settings/imported-addresses.jade
+++ b/app/partials/settings/imported-addresses.jade
@@ -23,7 +23,6 @@ h2(translate="SPENDABLE_ADDRESSES")
         th(translate="BALANCE")
         th
         th
-        th
       tr(
         imported-address="address", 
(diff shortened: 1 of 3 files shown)


assets/css/modules/_settings.scss
```diff
@@ -44,6 +44,28 @@
       &:hover { background: $hover-blue; }
     }
   }
+  // Accounts & Addresses
```
plondon: screen shot 2016-01-22 at 4 20 58 pm

app/partials/settings/imported-addresses.jade
```diff
@@ -23,7 +23,6 @@ h2(translate="SPENDABLE_ADDRESSES")
         th(translate="BALANCE")
         th
         th
-        th
```
plondon: screen shot 2016-01-22 at 5 08 15 pm

ROMANUCH05:
#267

PR #280, opened 2016-01-26
feat(dev): automatically login and go to last url when flag is set

jtormey:
A neat little dev feature I threw together. If the AUTO_RELOAD environment flag is set, the wallet will automatically log you in after a refresh and send you back to the last url you were at.


mpfluger:
Fancy

Sjors:
Fancy indeed. 

Can you remove the SAVE_PASSWORD feature, since that's already part of AUTO_RELOAD?

Also don't forget to update the README.

PR #281, opened 2016-01-27
Bcasyncform ui

plondon:


Sjors:
In the end it seems only file has changed. Can you combine these commits into one? E.g.:
`git reset HEAD~5` (goes back to the commit before you started, but leaves changes locally)
`git stash` (puts local changes on the stash)
`git reset --hard` (not sure if needed)
`git stash pop`
`git commit -a -m "fix(bcAsyncInput):..."` (new combined commit)
`git push --force`

PR #285, opened 2016-01-27
fix(alerts): use directive to display alerts, fix closing behavior

jtormey:
* Simplify alerts logic by putting it in a directive
* Fixed alerts so that they actually close when told
* Send modal will not display infinite alerts anymore


Sjors:
Very nice!

PR #286, opened 2016-01-27
fix(translations): fix display of translations followed by :

jtormey:


Sjors:
I wonder if coverall includes jade, css and json files in the coverage metric.

PR #289, opened 2016-01-27
fix(security): escape html where ng-bind-html is used

jtormey:
@kristovatlas


kristovatlas:
Manual code review looks good. Future note: `escapeHtml` is not a general-purpose XSS filter.

PR #296, opened 2016-02-02
Marketing page aka coolest landing page out there

plondon:


Sjors:
Looks sweet!

ACK, except for:
- don't copy paste the header bar code. Instead either:
 - use public.jade as the parent view of the marketing page, tweak where `ui-view="contents"` is placed so it works with login, signup and landing; or
 - create `public_header.jade` and put the reusable bits from `public.jade` there
- don't forget to put strings like "The New Blockchain Wallet" in `locales/en-human.json`

On the login and signup page, it might be nice if the top-left button takes you back to the landing page.

I also tested with `grunt dist`.

PR #297, opened 2016-02-03
Process tx

jtormey:


assets/js/directives/activity-feed.directive.js
```diff
@@ -23,11 +23,8 @@ function activityFeed($http, Wallet, Activity) {
       scope.activities = activities;
     });
 
-    scope.$watch('status.didLoadTransactions', (didLoad) => {
-      if (didLoad) {
-        Activity.updateTxActivities();
-        scope.loading = false;
-      }
+    scope.$watch(() => Activity.transactions, (txs) => {
+      if (txs.length > 0) scope.loading = false;
```
Sjors: And what if there's no transactions?

PR #307, opened 2016-02-10
Let users default to legacy address in receive

plondon:
screen shot 2016-02-10 at 1 02 18 pm

Opens ->

screen shot 2016-02-10 at 1 02 29 pm


plondon:
Made a small change to the flow, let me know when you get around to merging this. @jtormey

assets/js/controllers/request.controller.js
```diff
@@ -2,7 +2,7 @@ angular
   .module('walletApp')
   .controller("RequestCtrl", RequestCtrl);
 
-function RequestCtrl($scope, Wallet, Alerts, currency, $uibModalInstance, $log, destination, focus, $translate, $stateParams, filterFilter, $filter, format) {
+function RequestCtrl($rootScope, $scope, Wallet, Alerts, currency, $uibModalInstance, $log, destination, focus, hasLegacyAddress, $translate, $stateParams, filterFilter, $filter, format) {
```
Sjors: `hasLegacyAddress` needs to be present everywhere this modal is opened. That was missing in the addresses screen, so the modal broke. Fixed in c37a58032f043.

PR #308, opened 2016-02-10
Match header alignment to .info

plondon:
screen shot 2016-02-10 at 4 17 01 pm


plondon:
Waiting on .info logo update

PR #321, opened 2016-02-22
Dynamic fee

jtormey:
Todo:
- [x] dynamic URL
- [x] more tests for send.controller
- [ ] tests for DynamicFeeController

Depends on: 
- [x] Blockchain/My-Wallet-V3#117
- [x] https on endpoint

### DIFF
--- a/app/partials/dynamic-fee.jade
+++ b/app/partials/dynamic-fee.jade
@@ -1,17 +1,24 @@
 .modal-header.flex-row.flex-center
-  h3(ng-show="currentFee < suggestedFee" translate="LOW_FEE_WARN")
-  h3(ng-show="currentFee > suggestedFee" translate="HIGH_FEE_WARN")
+  div(ng-hide="surge")
+    h3(ng-show="currentFee < suggestedFee" translate="LOW_FEE_WARN")
+    h3(ng-show="currentFee > suggestedFee" translate="HIGH_FEE_WARN")
+  div(ng-show="surge")
+    h3(translate="SURGE_WARN")
   helper-button(content="DYNAMIC_FEE_HELP" linktext="LEARN_MORE" link="https://en.bitcoin.it/wiki/Transaction_fees")
 .modal-body
-  .col-sm-12
+  .col-sm-12(ng-hide="surge")
     p(ng-show="currentFee < suggestedFee" translate="LOW_FEE_EXPLAIN" translate-values="{ fee: currentFee, suggested: suggestedFee }")
     p(ng-show="currentFee > suggestedFee" translate="HIGH_FEE_EXPLAIN" translate-values="{ fee: currentFee, suggested: suggestedFee }")
+  .col-sm-12(ng-show="surge")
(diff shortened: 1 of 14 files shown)


tests/controllers/send_ctrl_spec.coffee
```diff
@@ -66,6 +77,14 @@ describe "SendCtrl", ->
 
     return
 
+  beforeEach ->
+    $httpBackend.expectGET('http://service-dynamic-fee.prod.blockchain.co.uk/fees')
```
Sjors: This needs to be dynamic, if only because production is HTTPS. Easiest way is to use grunt magic similar to how $rootScope.rootURL is set. I'll look into that.

Sjors: Well, not the test of course, but the original.

assets/js/controllers/send.controller.js
```diff
@@ -32,14 +33,25 @@ function SendCtrl($scope, $log, Wallet, Alerts, currency, $uibModalInstance, $ti
...
     destinations: [null],
     amounts: [null],
-    fee: Wallet.settings.feePerKB,
+    fee: 10000,
     note: "",
     publicNote: false
   };
 
-  $scope.payment = new Wallet.payment();
+  $scope.payment = new Wallet.payment({ feePerKb: 30000 });
```
Sjors: Why this default? Is that to prevent max balance issues?

app/partials/dynamic-fee.jade
```diff
@@ -0,0 +1,24 @@
+.modal-header.flex-row.flex-center
+  div(ng-hide="surge")
+    h3(ng-show="currentFee < suggestedFee" translate="LOW_FEE_WARN")
+    h3(ng-show="currentFee > suggestedFee" translate="HIGH_FEE_WARN")
+  div(ng-show="surge")
+    h3(translate="SURGE_WARN")
+  helper-button(content="DYNAMIC_FEE_HELP" linktext="LEARN_MORE" link="https://en.bitcoin.it/wiki/Transaction_fees")
+.modal-body
+  .col-sm-12(ng-hide="surge")
+    p(ng-show="currentFee < suggestedFee" translate="LOW_FEE_EXPLAIN" translate-values="{ fee: currentFee, suggested: suggestedFee }")
```
Sjors: Maybe make this `minFeeThreshold` and `maxFeeThreshold` so it doesn't immediately complain if someone rounds the number up or down?

Sjors: We could also use the 1 block fee as maximum and the within 6 blocks fee as minimum.

assets/js/controllers/send.controller.js
jtormey: 30 satoshi per byte is a reasonable fee setting, it's there in case the dynamic fee service fails in some way.

jtormey:
@Sjors can you look into making the fee service url dynamic? Once that is done, I think this is good to merge.

Sjors:
I'll do that.

PR #323, opened 2016-02-22
Remove recovery controller and tests

plondon:


Sjors:
Looks like orphaned code to me as well.

PR #326, opened 2016-02-25
Contextual Security Warning

jtormey:
Todo:

- [x] at least 5 tests

### DIFF
--- a/assets/js/directives/contextual-message.directive.js
+++ b/assets/js/directives/contextual-message.directive.js
@@ -42,8 +42,9 @@ function contextualMessage($cookies, $window, Wallet, SecurityCenter, filterFilt
       let didBackup     = Wallet.status.didConfirmRecoveryPhrase;
       let has2FA        = Wallet.settings.needs2FA;
       let verifiedEmail = Wallet.user.isEmailVerified;
+      let isSecure      = didBackup && (has2FA || verifiedEmail);
 
-      return isTime && hasBalance && !didBackup && !(has2FA || verifiedEmail);
+      return isTime && hasBalance && !isSecure;
     };
 
     scope.revealMsg = () =>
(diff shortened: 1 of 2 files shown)


assets/js/directives/contextual-message.directive.js
```diff
@@ -37,10 +37,13 @@ function contextualMessage($cookies, $window, Wallet, SecurityCenter, filterFilt
     };
 
     scope.shouldShow = () => {
-      let balance   = Wallet.total('');
-      let security  = SecurityCenter.security.score;
-      let isTime    = scope.msgCookie ? Date.now() > scope.msgCookie.when : true;
-      return balance > 20000000 && security < 0.5 && isTime;
+      let isTime = scope.msgCookie ? Date.now() > scope.msgCookie.when : true;
```
Sjors: What does scope.msgCookie do?

assets/js/directives/contextual-message.directive.js
```diff
@@ -37,10 +37,13 @@ function contextualMessage($cookies, $window, Wallet, SecurityCenter, filterFilt
...
+      let hasBalance    = Wallet.total('') > 0;
```
Sjors: Not that this can be null. Probably should make a test case for that just in case we change the rules and forget about the null case.

assets/js/directives/contextual-message.directive.js
```diff
@@ -37,10 +37,13 @@ function contextualMessage($cookies, $window, Wallet, SecurityCenter, filterFilt
...
+      let didBackup     = Wallet.status.didConfirmRecoveryPhrase;
+      let has2FA        = Wallet.settings.needs2FA;
+      let verifiedEmail = Wallet.user.isEmailVerified;
+
+      return isTime && hasBalance && !didBackup && !(has2FA || verifiedEmail);
```
Sjors: has2FA is can be null / undefined while loading this data. Balance is loaded separately. We should be careful not to assume a load order. Currently && means this won't be a problem, but a few test cases can make sure we don't break it later.

assets/js/directives/contextual-message.directive.js
jtormey: msgCookie is the data from the contextual-message cookie. This is how we check to see how much time has gone by since the security warning has been dismissed.

PR #335, opened 2016-03-02
Dynamic fee

plondon:


assets/js/controllers/send.controller.js
```diff
@@ -481,22 +481,24 @@ function SendCtrl($scope, $log, Wallet, Alerts, currency, $uibModalInstance, $ti
...
+        if (closestBlock === low && fee < Math.floor(low)) {
+          $scope.blockIdx = 6
+        } else if ( closestBlock === low ) {
+          $scope.blockIdx = 5
         }
 
-        $scope.confirmationTime = (blockIdx + 1) * 10
-        $scope.blockQueue = (blockIdx + 1)
-      }, 10)
+        $scope.confirmationTime = ($scope.blockIdx + 1) * 10
```
Sjors: Might be worth moving these calculations into a service.

app/partials/send.jade
```diff
@@ -118,14 +118,16 @@
...
-          ng-pluralize(count="blockQueue" 
-                       ng-init="block='BLOCK'; blocks='BLOCKS'"
-                       when="{'1': '{{block | translate}}', 'other': '{{blocks | translate}}' }")
-          span ).
-      //- QR Scanner
+        span.help-block.friendly(ng-show="!sendForm.$invalid && !hasInsufficientError(0) && confirmationTime && blockQueue" ng-class="{'has-error': blockIdx >= 6}")
+          span(ng-show="blockIdx <= 5")
+            span(translate="ESTIMATED_TX_TIME" translate-values="{ confirmationTime: confirmationTime, blockQueue: blockQueue }")
+            |  
+            ng-pluralize(count="blockQueue"
```
Sjors: Might be worth moving some of this complexity into one or more directives.

Sjors:
utACK

Sjors:
Can you move the dynamic fee stuff into a service? That way you can do:

```js
$timeout(() => $scope.blockIdx = Fee.blockIdx($scope.transaction.fee, ...)
```

That in turn means you can write tests for `Fee.blockIdx` to check that it returns 6, 7, etc. 

`guessAbsoluteFee` should also go to that that service.

The service should also handle the $http work.

Ideally the controller just does:
`$scope.surgeWarning = Fee.surgeWarning()`

`Fee.surgeWarning()` (and similar functions) return a promise and the service calls the endpoint if it hasn't already. A slightly lazier alternative is `Fee.initialize()` which returns a promise and then you call `Fee.surgeWarning`.

This makes is much easier to test those methods and should lead to a nett increase in test coverage.

Sjors:
Looks better. There's still room to move more functions out of the controller and into the feeService, but my understanding is that @pernas is moving it all the way to Blockchain/My-Wallet-V3.

PR #338, opened 2016-03-04
Display correct available balance in advanced send

jtormey:


Sjors:
I need either tests or a bunch of before-after screenshots; I can't tell what this fixes and what could go wrong.

PR #343, opened 2016-03-08
Move QR scanner to popover window

jtormey:
@plondon we talked about adding something like this a few weeks ago, so I took a few minutes to put this together. All the functionality is there, but it might need some design help (in advanced send in particular).

- [x] fix QR processing in send when amount is present.
- [x] fix advanced send overflow issues (phil).
- [x] test for onAddressScan

qr-scan-popup

### DIFF
--- a/app/partials/send.jade
+++ b/app/partials/send.jade
@@ -51,7 +51,8 @@
                 name="{{ 'destinations' + $index }}"
                 ng-model="transaction.destinations[$index]"
                 ng-model-options="{ updateOn: 'blur' }"
-                ng-change="updateToLabel(); checkForSameDestination(); setQRIndex($index)")
+                ng-change="updateToLabel(); checkForSameDestination();"
+                on-payment-request="applyPaymentRequest(request, $index)")
             .has-error
               span.help-block(translate="BITCOIN_ADDRESS_INVALID" ng-show="sendForm['destinations' + $index].$error.isValidAddress && sendForm['destinations' + $index].$touched && transaction.destinations[$index].address !== ''")
               span.help-block(translate="SAME_DESTINATION" ng-show="sendForm['destinations' + $index].$error.isNotEqual")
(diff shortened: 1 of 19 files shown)


Sjors:
Looks pretty sweet and I like how it removes complexity from the send controller and view.

Can you add a test for onAddressScan?

app/partials/send.jade
```diff
@@ -51,8 +51,7 @@
                 name="{{ 'destinations' + $index }}"
                 ng-model="transaction.destinations[$index]"
                 ng-model-options="{ updateOn: 'blur' }"
-                ng-change="updateToLabel(); checkForSameDestination(); refreshTxProposal()"
-                on-request-qr="cameraOn($index)")
+                ng-change="updateToLabel(); checkForSameDestination(); refreshTxProposal()")
```
plondon: It doesn't look like we use refreshTxProposal() anywhere

app/partials/send.jade
```diff
@@ -128,11 +127,6 @@
                          when="{'1': '{{block | translate}}', 'other': '{{blocks | translate}}' }")
             span ).
           span.has-error(ng-show="blockIdx > 6" translate="SMALL_FEE_WARNING")
-      //- QR Scanner 
-      .form-group.row(ng-hide="!cameraIsOn")
-        label.col-sm-2.col-xs-12(translate="QR")
-        .col-sm-10.col-xs-12
-          bc-qr-reader(active="cameraRequested", on-result="processURLfromQR" on-error="onError" camera-status="cameraIsOn")
```
plondon: removing processURLfromQR call also removes call to which removes the call to applyPaymentRequest, so if you scan a QR with an amount attached the amount field won't update.

plondon:
looks awesome!

- added it to import-address as well and left a few comments re. processing the QR in send.
- the loading in and fading out are a little sketchy, but I can try and fix that.

jtormey:
Great, thanks!

Sjors:
If you to advanced send, add one row, scan an address QR code with the 2nd row, then scan the an address+amount QR code with the 1st row, the amount will be added to the 2nd row.

Also if you scan an address on the 2nd row and then scan another address on the 1st row, both rows will have the new address.

PR #347, opened 2016-03-10
Use Node 5.10 and updated README

Sjors:
See also blockchain/My-Wallet-V3#171

I'm not seeing any negative effects in the application.


Sjors:
Half the tests are broken after I upgraded karma, phantoms, grunt-karma, etc.

The tests seem to be interfering with each other; e.g. all directive tests pass as long as you don't run controller tests.

Sjors:
I had to disable most of the coverage, because it's throwing: `[TypeError: Cannot read property 'text' of undefined]` 

I think that will work again after the tests are repaired.

Sjors:
I had to add `build/js/polyfill.js` to `karma.conf.js` and I had to disable one controller test, because that function sets window.location, which in turn lead to all directive tests failing with `SecurityError: SecurityError: DOM Exception 18`.

Sjors:
Yippee kayee, ******

PR #357, opened 2016-03-16
Transaction List Performance

jtormey:
Gets rid of any noticeable lag when navigating between / in transaction lists by:
* Adding one-way binding to expressions that don't need reevaluation to reduce watcher count
* Gradually rendering transactions on scroll
* Creating a filteredTransactions alias to prevent unnecessary list traversals


Sjors:
ACK

A lot of magic happens around `filteredTransactions` which is hard to test because it's in the Jade files. It might be better to do some of that in the controller, but I don't know if that plays nicely with ng-repeat.

PR #358, opened 2016-03-17
Marketing optimization

plondon:
Any files with a label TEMP, or code with comments surrounded with TEMP will be removed in the next iteration of the marketing page. It looks like a lot of lines of code have been added but most of them are from SVG files.

todo

- [x] add tests


assets/js/directives/scroll-in-view.directive.js
```diff
@@ -9,6 +9,7 @@ function scrollInView($window, $timeout) {
   const directive = {
     restrict: 'E',
     templateUrl: 'templates/scroll-in-view.jade',
+    replace: true,
```
jtormey: Why does this need to replace?

jtormey:
ACK

assets/js/directives/scroll-in-view.directive.js
plondon: It seemed like have <scroll-in-view><div class="scroll-in-view"> was a unnecessary.

app/index.jade
```diff
@@ -176,9 +179,27 @@ head
...
   <!-- @endif -->
+
+body(ng-controller="AppCtrl", ui-view="body")
   
+  //- This does not depend on Angular and will show as soon as the images are done loading.
+  <!-- @if PRODUCTION !>
+  .blocket-loading
+    div
+      img.full(src="/img/blocket-full-4847e61888d2d3c9f33f14840f92783d1c91dea2.png")
+      img(src="/img/blocket-no-boosters-c3f2a027ddfdbbce8eba8fc5da84c65d89d14fbc.png")
```
Sjors: No need to disable this in non-production. Also no need to put the sha hash in the image url's. This is handled by `grunt dist_unsafe` / `grunt release`.

app/index.jade
```diff
@@ -176,9 +179,27 @@ head
...
+  <!-- @endif -->  
+
+  noscript
+    img(src="/img/logo-bw.png")
+    h3 Please enable JavaScript to use the Blockchain Wallet
+
+  //- Load application.min.js in the body async.
+  <!-- @if PRODUCTION !>
+  script(src='js/application.min.js')
```
Sjors: If you want to load javascript in the body on production, you should do the same in dev. I'll add a commit.

Sjors: Turns out that there are side-effects when loading Javascript in the body. For instance `$rootScope.rootURL =` in `wallet.service.js` gets called before ng-init has set the value.

cc @jtormey

jtormey: Interesting, good to know. I'll see how we can deal with that.

PR #360, opened 2016-03-18
Marketing page

plondon:


Gruntfile.coffee
```diff
@@ -115,7 +115,10 @@ module.exports = (grunt) ->
 
     autoprefixer: {
       options: {
-        browsers: ['last 2 versions']
+        browsers: ['last 2 versions'],
+        # TEMP
+        remove: false
+        # END TEMP
```
plondon: We can remove this after we get rid of Bootstrap 2.0.4

PR #371, opened 2016-04-01
Sweep all imported addresses

jtormey:


locales/en-human.json
```diff
@@ -290,6 +290,11 @@
   "VERIFY_EMAIL" : "Verify your email address",
   "RESEND_EMAIL_VERIFICATION" : "Resend Verification Email",
   "SWEEP" : "Sweep",
+  "SWEEP_ALL" : "Sweep All",
+  "SWEEPING" : "Sweeping",
+  "SWEEP_WARNING" : "This will sweep all of your active addresses into your default account. This action cannot be undone. Addresses will be archived after being swept.",
```
plondon: This sounds like I'm sweeping all of my addresses into my default account, instead of just my imported addresses. Also can we say what the default account is? Like: "...into your default account (My Bitcoin Wallet). This action cannot be undone..."

assets/js/controllers/settings/sweepAll.controller.js
```diff
@@ -0,0 +1,33 @@
+angular
+  .module('walletApp')
+  .controller('SweepAllController', SweepAllController);
+
+function SweepAllController($scope, $q, $timeout, $translate, Wallet, Alerts) {
+  $scope.alerts = [];
+  $scope.defaultIdx = Wallet.my.wallet.hdwallet.defaultAccountIndex;
+  $scope.addresses = Wallet.legacyAddresses().filter(a => a.active && !a.isWatchOnly);
```
plondon: It might be helpful to display these addresses and their amounts in the first step of the Sweep All modal. Something like: 
screen shot 2016-04-01 at 11 35 55 am

tuzzolo: Should also add language as to whether or not the sweep will happen in one big transaction or one transaction per address.

Sjors: The latter would be better privacy wise.

plondon:
Question more than comment. I have a bunch of imported addresses with 0 BTC. When I ran Sweep all those addresses were "included" in the sweep, but they remained in imported addresses afterwards, is this because they're empty? What other reason is there for "Some addresses could not be swept automatically"? In the first step we should show only the addresses we are going to be able to sweep and give more helpful error messaging at the end if they weren't swept.

jtormey:
Some addresses might have a balance, but be made up entirely of very small coins. In this case, the address can't be swept because the coins are essentially worthless. The only way to spend then would be to go to advanced send and set the fee yourself.

Sjors:
nACK (apart from the tests)

Sjors:
utACK (apart from the tests)

jtormey:
Feature branch moved to #398

PR #374, opened 2016-04-04
Send fixes

jtormey:


app/partials/send.jade
```diff
@@ -116,7 +116,7 @@
                   name="fee"
                   tabindex="1"
                   ng-model="transaction.fee"
-                  min="0"
+                  min="1"
```
plondon: I think I changed this because chrome was showing a helper text: "Please enter a valid value. The two nearest valid values are 0 and 1" if a decimal was entered.

jtormey: Strange, novalidate is turned on so that shouldn't be an issue.

PR #385, opened 2016-04-07
Angular uib upgrade to 1.3.*

plondon:


Sjors:
I released a [new version of bc-phone-number](https://github.com/blockchain/bc-phone-number/releases/tag/v5.0.11) that allows for ui.bootstrap 1.3, so tests should now pass.

PR #391, opened 2016-04-18
Accessibility

plondon:
accessibility quick wins

### DIFF
--- a/app/index.jade
+++ b/app/index.jade
@@ -3,7 +3,7 @@ doctype
 <!-- @if PRODUCTION !><html lang='en' ng-app='walletApp' ng-csp ng-class="{'not-fixed': outOfApp}"><!-- @endif -->
 head
   meta(charset='utf-8')
-  meta(name="viewport",content="width=device-width, initial-scale=1.0, maximum-scale=1") 
+  meta(name="viewport",content="width=device-width, initial-scale=1.0") 
   meta(name="theme-color" content="#187fc0")
   meta(name="apple-itunes-app" content="app-id=493253309")
   link(rel="apple-touch-icon" sizes="180x180" href="img/favicon/apple-icon-180x180.png")
--- a/app/index.jade
+++ b/app/index.jade
@@ -42,8 +42,8 @@ body(ng-controller="AppCtrl" ui-view="body" ng-mousemove="resetInactivityTime()"
   //- This does not depend on Angular and will show as soon as the images are done loading.
   .blocket-loading
     div
-      img.full(src="img/blocket-full.png")
-      img(src="img/blocket-no-boosters.png")
+      img.full(src="img/blocket-full.png" alt="Loading")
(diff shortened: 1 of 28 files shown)


Sjors:
Please rebase this to master (semi standard). Also if you know any quick way to get test coverage to positive that would be great, but not very important.

app/partials/navigation.jade
```diff
@@ -12,7 +12,7 @@
     .navbar-collapse.collapse(ng-class="{'in bg-blue' : !navCollapsed}", ng-click="navCollapsed=true")
       ul.nav.navbar-nav.navbar-right(ng-show="status.isLoggedIn")
         li.item(ng-show="settings.btcCurrency")
-          a.flex-important.flex-column(ng-click="openBCmarkets()")
+          a.flex-important.flex-column(ng-click="openBCmarkets()" href="")
```
Sjors: Are you sure it's a good idea to put `href=""` everywhere? Doesn't AngularJS do that? Or is there some way to make AngularJS do it automatically?

app/index.jade
```diff
@@ -42,8 +42,8 @@ body(ng-controller="AppCtrl" ui-view="body" ng-mousemove="resetInactivityTime()"
   //- This does not depend on Angular and will show as soon as the images are done loading.
   .blocket-loading
     div
-      img.full(src="img/blocket-full.png")
-      img(src="img/blocket-no-boosters.png")
+      img.full(src="img/blocket-full.png" alt="Loading")
```
Sjors: Can you use angular translate here? Looks like it's mostly strings we've already localized, plus a small number of new ones.

plondon: Since we're loading this before Angular doesn't that stop us from being able to us translate here?

app/partials/navigation.jade
plondon: href="" makes the anchor tag focusable, which allows users to tab around and select. Without href="" the anchor tag is skipped over. I'm not sure if angular can put href="" everywhere, but I only added them to places I thought were important, like the dropdown menu in the header, or the dropdown for "accounts" in send.

app/index.jade
Sjors: Wrong example I guess. Inside index.jade we can't use AngularJS magic, so this line should just be in English.

plondon: You're right, just noticed the other examples of alt. I will update those, thanks

app/partials/navigation.jade
Sjors: Ah I see, if it's in specific places then this makes sense. Though it might be more clear for other developers is you add a directive like `selectable`.

plondon: True, random href=""'s dont make much sense out of context.

Sjors:
Can you rebase this?

plondon:
This has so many conflicts I'm just going to redo it.

PR #392, opened 2016-04-20
Proper favicon production URL's

Sjors:
The current production version uses `img/favicon.ico` which is renamed to `img/favicon-de0edd817019bf4d891472a6980e1a18cc19e927.ico` by `grunt dist`. 

The new fancy favicon folder that supports multiple browsers is in `img/favicon/*` which does not survive `grunt dist`. I fixed this, but I skipped `manifest.json`. 

There's two problems if you do want to use `manifest.json`:

1 - Grunt needs to process the image files first and then - in a seperate task - process `manifest.json`; this is because the contents of manifest.json depends on the new image file names.

2 - manifest.json uses `/` in the image URLs (which should fail even in dev mode) and it escapes slashes, which would mean we have to customize the grunt rename task to deal with those escape characters.

It would be easier to put Android favicons directly inside index.html if that's possible.

`browserconfig.xml` was already skipped in your version.


plondon:
I don't think manifest.json is actually necessary since android should pick up https://github.com/blockchain/My-Wallet-V3-Frontend/blob/master/app/index.jade#L18, so I'll remove it for now.

Sjors:
How would Android pickup e.g. `android-icon-36x36-86b53222bd2bfde43bf6be8f728b257eb435ead6`? There's no reference to it in index.jade.

plondon:
Where are you seeing `android-icon-36x36-86b53222bd2bfde43bf6be8f728b257eb435ead6`? I see `img/favicon/android-icon-192x192-61b0cd94cd3e0919b206e2f92a5af7c9ddcef1b6.png` in index.html after the build. Android should pick that up.

Sjors:
In `/dist` after you run `make dist`.

PR #393, opened 2016-04-21
Ui cleanup

plondon:
For buttons let's continue using the helpful button mixin, and omit any button-lg usage.

.button-success represents a final action (submit, save, create, update) btw these are harder to find, i'll be changing them as we go
.button-primary represents actionable buttons that lead to somewhere else
.button-danger represents canceling or warning buttons

todo
- [x] better .well class
- [x] send overview
- [x] remove transfer() from security
- [x] move transfer to more options in imported addresses


plondon:
@jtormey doesn't look like we're using https://github.com/blockchain/My-Wallet-V3-Frontend/blob/master/assets/js/controllers/securityCenter.controller.js#L24 anywhere. Can I remove it?

Sjors:
Please rebase this to master (semistandard).

PR #395, opened 2016-04-25
Semistandard

Sjors:
Make all Javascript [semistandard](https://github.com/Flet/semistandard) compliant.

Travis will fail if that's not the case. It's the developers responsibility to check this before making a pull request. This can be done by configuring the IDE or running `semistandard`. We could optionally add a grunt task for this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,10 @@ language: node_js
 node_js:
 - 5.10
 before_install:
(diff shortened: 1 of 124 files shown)


Sjors:
I removed the `don't merge yet` label because Travis will fail until this project is complete :-)

PR #400, opened 2016-04-26
Update backup recovery phrase title and descriptions

larskrueger:
screen shot 2016-04-26 at 16 56 38


Sjors:
For your next PR, can you name the commit like so:

`feat(Copy): update backup recovery phrase descriptions`

That way it gets picked up by the change log. You can also use `fix` instead of `feat` if the original was wrong, e.g. with a typo.

PR #401, opened 2016-04-26
fix(Gruntfile): forgot to rename file, causes IE 11 to miss Symbol

Sjors:


Sjors:
Don't delete the branch yet.

PR #404, opened 2016-04-28
Bug fixes

jtormey:


assets/js/controllers/app.controller.js
```diff
@@ -69,9 +69,8 @@ function AppCtrl ($scope, Wallet, Alerts, $state, $rootScope, $cookies, $locatio
   };
 
   $scope.$on('$stateChangeStart', (event, toState, toParams, fromState, fromParams) => {
-    if (toState.name === 'welcome' && Wallet.status.isLoggedIn) {
-      event.preventDefault();
-    }
+    let isPublicState = toState.name === 'welcome' || toState.name.slice(0, 6) === 'public';
```
Sjors: Probably good idea to check if `toState` and `toState.name` are defined.

PR #409, opened 2016-04-28
Error message when call to /unspent fails

jtormey:
Depends on [My-Wallet-V3 194](https://github.com/blockchain/My-Wallet-V3/pull/194)

Preview (copy might need updating):

### DIFF
--- a/assets/js/controllers/send.controller.js
+++ b/assets/js/controllers/send.controller.js
@@ -2,7 +2,7 @@ angular
   .module('walletApp')
   .controller('SendCtrl', SendCtrl);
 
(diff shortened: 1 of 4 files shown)


assets/js/controllers/send.controller.js
```diff
@@ -79,6 +86,16 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModalInstance, $t
     $scope.sendForm.$setUntouched();
   };
 
+  $scope.reopenModal = () => {
```
Sjors: Is there a nicer way to do this?

jtormey: Not that I know of, will look into it though

PR #419, opened 2016-05-03
Move angular-password-entropy dependency into frontend

jtormey:
Adapted from repo owned by @pernas: https://github.com/pernas/angular-password-entropy

Easier to maintain if it's in the frontend codebase.


pernas:
ack

Sjors:
Easier to maintain, but not easier to reuse by other projects. I know at least one (closed source) project that uses this library.

pernas:
@sjors the problem with the actual approach is that the logic related to scoring passwords is hided inside this small project. And moreover we have it duplicated in my-wallet-v3 because iOS needs to use the same logic. So with this change we will have this logic centralized in my-wallet-v3 and it is not going to be a pain every time kristov requires a change.

PR #426, opened 2016-05-05
fix(PrivateKeyImport): change confirmation for private key import

plondon:
@jtormey is there a better way to handle this next step action? Also does this flow make better sense to you?


jtormey:
Makes sense to me

PR #433, opened 2016-05-10
Use All

jtormey:


Sjors:
Can you add some tests for this? Otherwise ACK.

PR #434, opened 2016-05-10
Split Javascript

Sjors:
`application.min.js` is about 2.1 MB. I moved the My-Wallet-V3 javascript into a seperate file blockchain.min.js which is not loaded on the marketing page. This brings us down to 1.3 MB. Extracting the phone number selection UI and QR scanner library should make a big difference. We could remove even more files that the landing page doesn't require.

The script load is triggered in the wallet service after a route change away from the landing page. It's achieved by adding a script tag at the bottom of the body. Note that a script tag inside a Jade template is ignored by AngularJS because of jQuery Lite limitations.

All My-Wallet-V3 services are now loaded as promises. A timer loop checks if window.Blockchain is present and once the javascript is loaded these promises are resolved.

Some UI elements are replaces by a spinner while the Blockchain javascript is loading. E.g. the login form can be shown immedidately but the Login button is hidden, because login won't work until MyWallet is done loading. @plondon you can customize this behavior of course; e.g. you could hide the entire screen or you could enable the login button, but delay the call to MyWallet.login until that's loaded.

Unfortunately dealing with these services as promises required some refactoring. Worse, it broke almost every test. Recovering those tests is quite tedious.

I'm not sure if this approach is the best.

- [ ] check load sequence for rootURL (etc) in dist mode
- [ ] extract QR scanner library (load when send modal is opened)
- [ ] extract phone number library (especially the regexes are huge, load when settings / security center are loaded)
- [ ] extract even more?
- [ ] repair  tests
- [ ] @plondon get rid of rocket and/or replace with something more subtle? It's especially annoying if you click on the legacy login link and then hit the back button


plondon:
I agree, let's kill the blocket loading on the marketing page. I'll come up with something more subtle.

Sjors:
Superseded by #447

PR #435, opened 2016-05-10
Export History

jtormey:
Depends on
 - [ ] Tests for ExportHistoryController


Sjors:
It's a bit picky about leading zeros in dates:
date

I'm getting "no history"
no history

Maybe it's some weird javascript date formatting thing?
end date

The mutliaddr and marketdata calls do return data.

jtormey:
Looks like you flipped the dates around. I made "Start Date" the most recent date, and "End Date" the date the export goes back to. Not sure which makes more sense. Also, the format is MM/dd/yyyy (🇺🇸), is this something that should be changed / localized?

Sjors:
I would assume "start date" refers to the oldest date, but as long as you make it clear...

yyyy-mm-dd (the Chinese way) might be the least confusing. Or perhaps [MomentJS](http://momentjs.com) can do the right thing depending on the browser locale.

emilybremner:
Tend to side with Sjors on start date being oldest date, just cause that's what I'm used to from other tx histories (banks, credit cards etc.) feels more intuitive.

PR #441, opened 2016-05-17
Include watch-only balances

jtormey:


Sjors:
@jtormey Travis seems really unhappy.

jtormey:
It's because of the semistandard version mismatch that was fixed in the tx-details branch. Will include the fix here as well.

PR #442, opened 2016-05-17
Show warning when payment object emits a message

jtormey:


Sjors:
Can you add a test for this and figure out why Travis broke?

jtormey:
Travis broke because he installs semistandard globally. Semistandard just hit v8.0.0, while our package.json still uses 7.*. So travis is using some new rules. Fixed this in tx-details branch, will apply the fix here as well

PR #445, opened 2016-06-01
Bitcoin url

plondon:


assets/js/controllers/request.controller.js
```diff
@@ -143,10 +144,19 @@ function RequestCtrl ($rootScope, $scope, Wallet, Alerts, currency, $uibModalIns
...
+      url += `amount=${ parseFloat($scope.fields.amount / 100000000) }`;
+      url += `&`;
+      url += `message=${ $scope.fields.label }`;
+    } else if ($scope.fields.amount > 0) {
+      url += `?`;
+      url += `amount=${ parseFloat($scope.fields.amount / 100000000) }`;
+    } else if ($scope.fields.label !== '') {
+      url += `?`;
+      url += `message=${ $scope.fields.label }`;
     }
```
Sjors: Can you use https://docs.angularjs.org/api/ng/service/$httpParamSerializer here?

Sjors: ACK other than the above.

plondon: $httpParamSerializer does not encode the url in accordance to the BIP. I think to keep our url structure interoperable we should use encodeURI().

PR #447, opened 2016-06-03
Landing page loads much faster

Sjors:
Work in progress. I'm splitting the app into a landing page AngularJS app and a wallet AngularJS app. The landing page app will reuse some elements from the main app (translations, ui.bootstrap, the rootUrl, header page and some directives).

I got the landing page with all fonts and images down to about 500 kb compressed and @plondon should be able to get this down further.

Todo:
- [x] use [ocLazyLoad](https://github.com/blockchain/My-Wallet-V3-Frontend/compare/lazy-load-demo) instead of two separate apps
- [x] fix navigate from login page back to landing page 
- [ ] loading indicator after clicking login / signup (and prevent blank screen)
- [x] loading wallet app when user navigates to login / signup
- [x] use plain old href instead of ui-router on landing page
- [x] Phil: remove unnecessary includes from assets/css/landing.css.scss
- [x] production asset pipeline
- [x] repair tests
- [ ] add tests
- [x] load QR and phone number library just in time if they take too long to load (might be separate PR)


plondon:
What do you think about having landing-imports and wallet-imports loading everything they both need and then unloading the landing-imports when you load wallet-imports? That brings landing.css down to 45kb. BTW, once we remove utilities/legacy we'll drop 15kb.

Sjors:
Why do you want to unload the landing page CSS? Does it negatively impact performance?

The downside of this last change is that it makes the main wallet CSS file larger, since both CSS files now contain all of Bootstrap. It will also be tricky to control the timing of when the landing CSS is safe to delete and limits our options for a smooth transition (the current version just makes the page blank while loading).

Sjors:
If I enable bandwidth throttle in the Chrome dev tools I see this during the loading process:
load

We can probably tweak the views and css to make that nicer, without involving rockets. Let's do that in a separate PR though.

Sjors:
Let's also move "loading indicator after clicking login / signup" to a new PR. It's definately useful because if you enable bandwidth throttling and then click on the login button, the UI just sits there and does nothing for a few seconds.

Sjors:
Two tests were breaking on Travis but not locally because Travis filesystem is case sensitive. See last commit.

PR #451, opened 2016-06-03
Receive dropdown performance

jtormey:


app/partials/request.jade
```diff
@@ -13,12 +13,12 @@
...
-        ui-select(ng-model="fields.to" ng-disabled="paymentRequest.amount > 0 || paymentRequest.label")
-          ui-select-match(placeholder="{{'SEARCH' | translate}}...")
+        ui-select(ng-model="fields.to" ng-disabled="paymentRequest.amount > 0 || paymentRequest.label" tagging)
+          ui-select-match(placeholder="{{::'SEARCH' | translate}}...")
             label-origin(origin="$select.selected")
-          ui-select-choices(repeat="destination in destinations | filter: getFilter($select.search)" group-by="'type'")
-            span
-              label-origin(origin="destination" highlight="$select.search")
+          ui-select-choices(repeat="destination in destinations | filter: {label:$select.search} | limitTo:destinationLimit" group-by="'type'")
+            span(in-view="$last && destination.type==='Imported Addresses' && increaseLimit()")
```
plondon: This doesn't seem to work as expected. I created a wallet with 150 accounts it cuts off at 50, see attached screen shot. You may be able to pass more options to in-view to help it figure out what it's viewport limits are, or just take out the limitTo for now. Same issue in send.
screen shot 2016-06-04 at 1 21 44 pm

jtormey: Good catch. It's supposed to increase the number of list items when it reaches the last imported address. I didn't think about the case where the user has >50 accounts and 0 addresses. Probably an edge case, but will commit a fix for it.

PR #455, opened 2016-06-07
Display sticky browser error message in form

Sjors:
- [x] depends on blockchain/My-Wallet-V3#219

chrome

Moved browser detection to a directive. The message no longer
disappears after a few seconds. It's displayed right inside the login /
signup / recovery form.

You can test this by using an old browser or by increasing the required
version.


plondon:
LGTM

PR #456, opened 2016-06-07
Forms

plondon:
Work in progress

- [ ] Tests
- [x] Security Forms
- [x] Preferences Forms
- [x] Security Center Forms
- [x] Handle enter in all forms
- [x] Autofocus correct input field
- [x] Remove the waiting img gif and replace with ladda when applicable


assets/js/directives/settings-form.directive.js
```diff
@@ -0,0 +1,37 @@
+angular
```
plondon: @jtormey I wrapped these forms in a directive that handles the controller toggling to get rid of a lot of code reuse.

PR #472, opened 2016-06-16
Bugs

plondon:


app/partials/send.jade
```diff
@@ -26,7 +26,6 @@
             img(src="img/spinner.gif")
           ui-select.send-from-dropdown(
             ng-model="transaction.from"
-            ng-model-options="{debounce:250}"
```
plondon: https://github.com/angular-ui/ui-select/issues/855

jtormey: Can we replace with refresh-delay?

PR #480, opened 2016-06-24
Update email verification status in-app

jtormey:
- [x] https://github.com/blockchain/My-Wallet-V3/pull/232

PR #483, opened 2016-06-27
Parse (some of) get-info result in My-Wallet-V3

Sjors:
- [x] depends on blockchain/My-Wallet-V3#234
- [x] rebase after #447 and #480 are merged


Sjors:
This can be merged, but we should release a new bower version first.

PR #488, opened 2016-06-29
Remove alias

jtormey:


assets/js/controllers/settings/info.controller.js
```diff
@@ -2,30 +2,29 @@ angular
...
+  $scope.pairingCode = null;
 
-  $scope.display = {
-    pairingCode: false
+  $scope.removeAlias = () => {
+    $scope.loading.alias = true;
+    Alerts.confirm('CONFIRM_DISABLE_ALIAS', { id: $scope.guid })
+      .then(Wallet.removeAlias)
+      .then(() => $scope.alias = Wallet.user.alias, () => {})
+      .then(() => $scope.loading.alias = false);
```
Sjors: Does `loading.alias = false` get called if `Wallet.removeAlias` fails?

assets/js/controllers/settings/info.controller.js
```diff
@@ -2,30 +2,29 @@ angular
...
-      Alerts.displayError('Failed to load pairing code.');
-      $scope.loading = false;
-    };
-    $scope.loading = true;
-    Wallet.makePairingCode(success, error);
+    $scope.loading.code = true;
+    let success = (code) => $scope.pairingCode = code;
+    let error = () => Alerts.displayError('SHOW_PAIRING_CODE_FAIL');
+    $q(Wallet.makePairingCode)
+      .then(success, error)
```
Sjors: Why `.then(success, error)` instead of `.then(success).catch(error)`?

assets/js/controllers/settings/info.controller.js
```diff
@@ -2,30 +2,29 @@ angular
...
+      .then(() => $scope.loading.code = false);
```
Sjors: Does `loading.alias = false` get called if `Wallet.makePairingCode` fails?

locales/en-human.json
```diff
@@ -261,11 +261,16 @@
   "HIDE_PAIRING_CODE"  : "Hide Pairing Code",
   "SHOW_PAIRING_CODE"  : "Show Pairing Code",
   "PAIRING_CODE_EXPLAIN": "Scan the code <i>(click on 'Show Pairing Code')</i> with your Blockchain Wallet (iOS or Android) for a seamless connection to your wallet.",
+  "SHOW_PAIRING_CODE_FAIL" : "Failed to load pairing code",
   "DOWNLOAD_THE": "Download the",
   "IOS_APP_HERE": "iOS app here",
   "AND_THE": "and the",
   "ANDROID_APP_HERE": "Android app here",
   "MOBILE" : "Mobile",
+  "CONFIRM_DISABLE_ALIAS" : "<i class='ti-alert security-red mrs'></i>Remove Wallet Alias?<b><br/><br/><small>Once your alias is removed, you will not be able to use it to log in. This cannot be undone. Make sure you write down your Wallet Identifier before continuing.<br/><br/>Wallet ID: <b>{{::id}}</b></small>",
```
Sjors: Can you split this? A human translator might accidentally break this.

tests/controllers/settings_info_ctrl_spec.coffee
```diff
@@ -1,57 +1,60 @@
...
-      Wallet.makePairingCode(success, error)
-      expect(scope.display.pairingCode).toBe(true)
+      scope.$digest()
+      expect(scope.pairingCode).toEqual("code")
+      expect(scope.loading.code).toEqual(false)
 
-    it "should show an error when failed", inject((Alerts) ->
-      success = () ->
-      error = () ->
+    it "should display an error when show failed", ->
```
Sjors: Maybe change to "should display an error _and disable spinner_ when show failed"

assets/js/controllers/settings/info.controller.js
jtormey: Yes, all errors are dealt with before that final `.then`, so `loading.alias` will always be set to `false` no matter what happens.

assets/js/controllers/settings/info.controller.js
jtormey: In this case, because it's more concise.

assets/js/controllers/settings/info.controller.js
jtormey: Yeah, because the prior `.then` will handle any errors. So that last `.then` will always run.

locales/en-human.json
jtormey: You're right, it's better to have this html in a partial. We have a bunch of warning alerts that follow this format, will clean them up and add to the PR.

PR #499, opened 2016-07-11
Send confirmation screen

plondon:
screen shot 2016-07-11 at 4 27 49 pm

### DIFF
--- a/app/partials/send.jade
+++ b/app/partials/send.jade
@@ -169,29 +169,32 @@
       .flex-column
         .flex-column
           span.type-lbl.mbs.underline(translate="TO")
-          .mls.mbs.flex-row.flex-center.center.flex-between(ng-repeat="address in getToLabels() track by $index")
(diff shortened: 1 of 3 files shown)


Sjors:
For us non-CSS gurus, can you explain in one paragraph roughly what was wrong and why this fixes it?

plondon:
IE has issues interpreting some flexbox properties. The offender here was flex: 1. The default value for this property according to the spec is flex: 0 1 auto. The values are flex: [flex-grow (how much can I grow relative to siblings)] [flex-shrink (how much can I shrink relative to siblings)] [flex-basis (default val in px em or %)]. So in webkit flex: 1 resolves to flex: 1 1 auto. But in IE it resolves to flex: 1 0 auto. Meaning it can shrink to 0 width, which is what happened. I'll look for other occurrences of this as well before QA catches them.

I changed the markup to use width: 50% for the accounts and tx fee label, and width: 50% for the values. Since IE is fine with flex-row and flex-between I left them to display the items inline.

PR #507, opened 2016-07-13
Translate alert message in template

jtormey:
In the latest version of angular-translate, `$translate()` will only resolve if it can find a matching translation string, otherwise it will reject. This is a change from the previous version we were on, which would resolve with the argument (even if it wasn't a translation string). This caused alerts to not display in cases where we don't have translation strings in place.

This commit makes it so the alert will display even if no translation string is found.


Sjors:
`"angular-translate": "2.11.*",`
So much for semantic versioning... Good catch!

Sjors:
We use `$translate(...).then()` in more places, are those affected as well?

jtormey:
Possibly, I'll check if there's anywhere else that might be broken.

Sjors:
Update: we updated the minor version, so our bad :-)

PR #508, opened 2016-07-19
Prompt for main pw when viewing backup with no second pw

jtormey:


Sjors:
utACK if you add a few more tests

PR #509, opened 2016-07-19
Address book in send

jtormey:
Also now using ui-select in send to dropdown.


plondon:
I can no longer send to my default wallet.

jtormey:
Nice catch, I added that watch expression back in.

PR #529, opened 2016-07-26
Stops invalid form submissions

c18js:
Fixes the problem where user could hit enter to submit invalid forms in the settings panel. SAFARI ONLY


plondon:
Awesome 😎!

PR #534, opened 2016-07-27
tx-notes no longer able to have one long word overlapping

c18js:
In tx feed, one can no-longer have a single, long word which covers other elements


assets/css/modules/_transaction-list.scss
```diff
@@ -105,6 +105,10 @@
   .feed-details {
     border-top: 1px solid $eee;
   }
+  .tx-note {
+    max-width: 100%;
```
plondon: For next time, there's a .mw class for this.

PR #540, opened 2016-07-30
Accept terms of service during recovery

larskrueger:


plondon:
Can you:

git checkout recover_tos
git rebase master
[resolve conflicts]
git add .
git rebase --continue
git push origin recover_tos -f

app/partials/recover-funds.jade
```diff
@@ -75,6 +75,14 @@
         span.help-block 
           div(ng-show="recoveryForm.confirmation.$touched")
             p(ng-show="recoveryForm.confirmation.$error.isValid" translate="NO_MATCH")
+    .flex-center.flex-end.mtm.mbl
+      input#agreement_accept.pull-right(
+        name="agreement"
+        type="checkbox"
+        ng-model="fields.acceptedAgreement"
```
plondon: This doesn't correspond to anything in the controller as far as I can tell.

Sjors: It may not have to because of the `required` flag. @jtormey?

jtormey: Right, the form won't be valid unless the checkbox is checked. So you just need to look for that and not worry about `fields.acceptedAgreement`.

larskrueger: This is where it's copied from: https://github.com/blockchain/My-Wallet-V3-Frontend/blob/460cd883e031face9a9a865df8945a25c37a0f80/app/partials/signup.jade#L63-L67

PR #561, opened 2016-08-08
feat(Left Navbar): can now toggle currency from nav

jstettner:


app/templates/fiat-or-btc.jade
```diff
@@ -1,3 +1,3 @@
-span
+span(ng-click="toggle()")
```
plondon: I'm not sure we want this for every instance of fiat-or-btc

PR #564, opened 2016-08-09
feat(Tx): allows users to sort tx list

jstettner:


plondon:
Few notes:

1. I don't think we should show this if there are 0 transactions.

2. Maybe we could show a loading icon or something since sometimes the sort takes a while?

3. I'm not sure about the UI. I think it should be attached to the [Search] field with a dropdown option.

Sjors:
Closing this for now, as it's too far out of sync.

PR #565, opened 2016-08-09
fix(Addresses): buttons and search-bar become column mobile

jstettner:


plondon:
Why hide it at all?

jstettner:
@plondon doesn't anymore

PR #577, opened 2016-08-24
Enclose CSV fields in quotes

jtormey:


assets/js/services/wallet.service.js
```diff
@@ -1117,7 +1117,7 @@ function Wallet ($http, $window, $timeout, $location, Alerts, MyWallet, MyBlockc
   wallet.exportHistory = (start, end, active) => {
     let json2csv = (json) => {
       let headers = Object.keys(json[0]);
-      let makeRow = (obj) => Object.keys(obj).map(key => obj[key]).join(',');
+      let makeRow = (obj) => `"${Object.keys(obj).map(key => obj[key]).join('","')}"`;
```
Sjors: So what happens if there's a quote in a transaction note? In other words, should we use a library to do this?

jtormey: Good point @Sjors. We could use a library like [json2csv](https://github.com/zemirco/json2csv) to take care of that. Another option is to use `JSON.stringify`, which will escape quotes for us.

PR #578, opened 2016-08-29
Cream

Sjors:


assets/js/services/country.service.js
```diff
@@ -0,0 +1,236 @@
+// http://data.okfn.org/data/core/country-list/r/data.json
```
Sjors: Can we pull this in via bower or NPM instead?

PR #579, opened 2016-08-29
Use My-Wallet-V3 service, add country to signup

Sjors:
Depends on blockchain/My-Wallet-V3#256


Sjors:
I'll try to fix those tests tomorrow...

Sjors:
I repaired the tests. I noticed that some of the controller tests relied on the real buy-sell service. I replaced that with a mock.

The code that determines which step the user is on still feels a bit brittle. I think it deserves it's own directive / component / class.

PR #591, opened 2016-08-31
fix(Limits): calculate limits correctly

plondon:


Sjors:
I like ambitious commit messages :-)

PR #604, opened 2016-09-02
Use integer cents / satoshi instead of floats

Sjors:
Depends on blockchain/My-Wallet-V3#265


Sjors:
These PR's are ready to merge now.

`Quote`  now has a use `getPaymentMethods()` function, which populates `quote.paymentMethods.card` and `quote.paymentMethods.bank` with the right fees. You can probably simplify `buy.controller.js` a bit now.

Sjors:
It's making infinite requests to the anonymous quote endpoint when you open the modal. I'll deal with that on master.

PR #609, opened 2016-09-05
Tx description year

plondon:


plondon:
@mmaton Thanks for the pull request. I changed it a bit to show the year if the transaction is not from the current year.

PR #611, opened 2016-09-06
Out amount

plondon:
https://github.com/blockchain/My-Wallet-V3/pull/268

Using outAmount solves the issue where the user completes a trade but the amount does not show.


Sjors:
That PR is merged, so now this depends on blockchain/My-Wallet-V3#265

PR #623, opened 2016-09-09
Use correct limit amount for kyc state in buy homepage

jtormey:


assets/js/components/kyc-status.component.js
```diff
@@ -18,7 +18,7 @@ angular
 
       this.getCardMax = () => {
         let symbol = this.limits.currency && this.limits.currency.symbol;
-        let amt = this.limits.card && this.limits.card.available;
+        let amt = this.limits.card && this.limits.card.max;
```
Sjors: While you're at it, these should be cent integers.

PR #628, opened 2016-09-12
Remove Second Password

Sjors:
Second password removal now happens in a modal. The user needs to reload their wallet afterwards. It's not a huge issue if they don't, but buy-sell won't load.


Sjors:
I should add that the modal UI needs some love:
remove

assets/js/controllers/wallet.controller.js
```diff
@@ -8,6 +8,7 @@ function WalletCtrl ($scope, $rootScope, Wallet, $uibModal, $timeout, Alerts, $i
   $scope.status = Wallet.status;
   $scope.settings = Wallet.settings;
   $rootScope.isMock = Wallet.isMock;
+  $rootScope.needsRefresh = false;
```
Sjors: That workaround is fine with me.

Sjors:
utACK

PR #640, opened 2016-09-14
Buy refactor

plondon:
I went a bit beyond just kyc states here, but this should resolve some of the issues we were seeing.

1. Better step naming
2. Removed unused attrs from buy-steps
3. Combine all isx steps into 'isx' and all formatted trade states into 'trade-formatted'


jtormey:
This is fantastic, thanks @plondon

PR #641, opened 2016-09-14
Prevent multiple buy modals because of fast clicking

jtormey:


Sjors:
This might be an example where using ui.router to open the modal would be useful.

PR #642, opened 2016-09-15
refactor(BankTransfer): bank transfer is a formatted trade

plondon:


Sjors:
More red than green always makes me happy.

PR #645, opened 2016-09-15
fix(Bank Transfer): remove binding to prop and val so trade can update

plondon:
Although this will never actually be an issue IRL because bank transfers take a day or more and I doubt people will leave this modal open for that long, it will help QA. Need to remove the bindings so that the actual tx values update when the trade goes from pending user action to pending coinify action.


Sjors:
There are some countries where bank transfers are extremely fast, so this may come out handy in the future.

PR #650, opened 2016-09-16
Handle floats properly in when formatting fiat for view

jtormey:


tests/services/currency.service.spec.js
```diff
@@ -191,6 +191,11 @@ describe('currency', () => {
       let formatted = currency.formatCurrencyForView(amount, currency.bitCurrencies[0], false);
       expect(formatted).toEqual('0.12345679');
     }));
+
+    it('should format amounts that javascript has trouble with', inject((currency) => {
```
Sjors: Nice

PR #652, opened 2016-09-16
Prevent quickstart input and buy modal amounts from disagreeing

jtormey:
transaction.fiat was slow to update in the html, so pressing enter right after entering an amount would cause the modal to open with the old amount. Putting the logic into the scope makes it more responsive.


plondon:
Nice

PR #653, opened 2016-09-16
Use separate receive address for each transfer tx

jtormey:
@pernas let me know if you think this change should go in payment.js instead.


Sjors:
@jtormey I think it should, because we want this to work in iOs as well at some point.

jtormey:
Not sure of the best way to move this to payment.js. We only want the receive index to increment in the case where we know we're doing payments consecutively, not all the time. Might be better to use this fix for now, unless you have something else in mind.

Sjors:
Ok, just make sure @woudini is in the loop.

PR #661, opened 2016-09-19
Use quote.getPaymentMethods

jtormey:


assets/js/controllers/buysell/buy.controller.js
```diff
@@ -81,24 +83,24 @@ function BuyCtrl ($scope, $filter, $q, MyWallet, Wallet, MyWalletHelpers, Alerts
     $scope.status.waiting = true;
 
     let success = (methods) => {
-      $scope.methods = {
-        card: methods.filter((m) => m.inMedium === 'card')[0],
-        bank: methods.filter((m) => m.inMedium === 'bank')[0]
-      };
+      $scope.methods = methods;
```
plondon: Right, good call

assets/js/controllers/buysell/buy.controller.js
```diff
@@ -133,45 +135,25 @@ function BuyCtrl ($scope, $filter, $q, MyWallet, Wallet, MyWalletHelpers, Alerts
...
-
-      return quote.getPaymentMethods().catch(getPaymentMethodsError);
-    };
-
-    const setQuote = (quote) => {
-      $scope.quote = quote;
-      return quote;
-    };
+    let quoteError = eventualError('ERROR_QUOTE_FETCH');
+    let amount = Math.round($scope.transaction.fiat * 100);
```
plondon: Do you think calling Math.round here and not calling it on 257 will give us problems later on?

jtormey: It might cause this assertion to fail at some point: https://github.com/blockchain/My-Wallet-V3/blob/master/src/coinify/coinify.js#L242

For example: `Math.round(35.05 * 100) === 35.05 * 100 // false`

jtormey: Nice find, that assertion was failing every time I entered 35.05, fixed now

Sjors:
Thanks, `buy.controller.js` is becoming more readable...

PR #662, opened 2016-09-19
refactor(Buy): folder structure and organization

plondon:


jtormey:
Good move

Sjors:
Nice, can you move the tests into their own folder as well?

Sjors:
Even if we never actually split it off, it would be nice to organise code such that it's possible to run standalone in any Angular application, e.g. as `ng-buy-sell-bitcoin`.

PR #663, opened 2016-09-19
feat(Download receipt): download receipt

plondon:


Sjors:
As much as I love puppies*, shouldn't we mark this "don't merge yet" until the API supports this?

* = I don't

plondon:
The API does support this, only for real purchased/completed trades.

PR #666, opened 2016-09-20
fix(Buy select country): when searching filter search by startsWith

plondon:


assets/js/controllers/buysell/buySelectCountry.controller.js
```diff
@@ -32,4 +32,12 @@ function BuySelectCountryCtrl ($scope, country, MyWallet, buySell) {
     let country = $scope.countries.countryCodes.filter(c => c['Code'] === $scope.fields.countryCode)[0]['Name'];
     buySell.signupForAccess(email, country);
   };
+
+  $scope.startsWith = (viewValue, search) => {
+    if (!search.length) return true;
+    else {
+      return viewValue.toLowerCase && viewValue.toLowerCase().startsWith(search.toLowerCase()) ||
+             viewValue.name && viewValue.name.toLowerCase().startsWith(search.toLowerCase());
```
jtormey: Does `startsWith` get transpiled by babel? Or do all of our supported browsers have this?

plondon: According to the docs this isn't supported by ie11 (although when I tested it it worked). I'll just use a different method instead of adding a shim.

PR #669, opened 2016-09-20
fix(Quick Start): update quick start quote every minute

plondon:


assets/js/directives/buy-quick-start.directive.js
```diff
@@ -36,24 +36,31 @@ function buyQuickStart (currency, buySell, Alerts) {
...
+      scope.quote = quote;
+      scope.status = {};
+      Alerts.clear();
+    };
+
+    const error = () => {
+      Alerts.displayError('ERROR_QUOTE_FETCH');
+    };
+
+    scope.$on('$destroy', () => stopFetchingQuote);
```
jtormey: `stopFetchingQuote` won't get called here

PR #674, opened 2016-09-21
Prevent blank page when refreshing on settings/:account/addresses

jtormey:
Angular UI Router would silently throw an error and freeze up, since it wasn't able to resolve Wallet.


Sjors:
I was wondering why refresh didn't work on that page, good catch.

PR #687, opened 2016-09-28
Receive: "copy next address" button

jtormey:
https://i.gyazo.com/13bbcdf3126ba2664d7eda673000af1e.gif

### DIFF
--- a/assets/js/directives/on-click-copy.directive.js
+++ b/assets/js/directives/on-click-copy.directive.js
@@ -1,0 +1,43 @@
+
+angular
+  .module('walletApp')
+  .directive('onClickCopy', onClickCopy);
+
+function onClickCopy ($window, $document) {
+  const directive = {
+    restrict: 'A',
+    scope: false,
+    link: link
+  };
+  return directive;
+
+  function link (scope, elem, attrs) {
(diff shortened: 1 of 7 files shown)


assets/js/directives/on-click-copy.directive.js
```diff
@@ -0,0 +1,43 @@
...
+    scope.copy = () => {
+      let text = scope.$eval(attrs.onClickCopy);
+
+      if (!angular.isString(text) || text.length === 0) {
+        return;
+      }
+
+      let textElem = angular.element(`<span>${text}</span>`);
+      textElem.css({ position: 'absolute', left: '-1000px' });
+      angular.element($document[0].body).append(textElem);
```
Sjors: Can you do this with a normal template instead?

jtormey: What do you mean by normal template? We need to create a dom node that's out of view, so that we can tell the browser to copy its text. Not sure if we can do that with a conventional directive.

Sjors: I see you're adding it to the end of the DOM, not inside the modal. That would be tricky indeed.

PR #688, opened 2016-09-29
Guess currency from country code

Sjors:
We'll need a new endpoint to get the users country code (based on IP).

### DIFF
--- a/assets/js/controllers/signup.controller.js
+++ b/assets/js/controllers/signup.controller.js
@@ -18,16 +18,120 @@ function SignupCtrl ($scope, $state, $cookies, $filter, $timeout, $translate, Wa
     language_guess = $filter('getByProperty')('code', 'en', languages);
   }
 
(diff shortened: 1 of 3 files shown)


assets/js/controllers/signup.controller.js
```diff
@@ -18,16 +18,120 @@ function SignupCtrl ($scope, $state, $cookies, $filter, $timeout, $translate, Wa
...
+      break;
+    case 'jp': // Japan
+      cur = 'JPY';
+      break;
+    case 'pl': // Poland
+      cur = 'PLN';
+      break;
+    case 'gb': // Great Britain, United Kingdom, Channel Islands
+    case 'im': // Isle of Man
+    case 'gg': // Guernsey
```
alecalve: Maybe add `je` for Jersey?

Sjors: Thanks, done.

PR #693, opened 2016-10-03
Language routing, dropdown and search in public nav

jtormey:


Sjors:
Are you sure the backend will support this language prefix when e.g. downloading the recovery prefix? cc @mpfluger 
pdf

There's a UI glitch if the browser isn't wide enough:
schermafbeelding 2016-10-04 om 11 17 51

Other than that it looks good to me.

plondon:
LGTM

jtormey:
@Sjors the backend is just forwarding /LC/wallet/* to /wallet/*, so I imagine everything works the same. I'll push a fix for that browser width issue.

@plondon yeah search won't work locally because of our dev server. I can add a redirect for local though.

mpfluger:
@sjors, yes, this will work ;)

PR #704, opened 2016-10-11
/buy instead of /buy-sell

Sjors:


jtormey:
Decision was made to leave it as is for now.

PR #708, opened 2016-10-11
Logout flow update

jtormey:
Depends on https://github.com/blockchain/My-Wallet-V3/pull/295
Session not terminated by logging out, instead user is asked to "Secure Wallet"


Sjors:
I see you did some spring cleaning too, nice!

PR #713, opened 2016-10-13
Backup / Transfer All

jtormey:


assets/js/controllers/confirmRecoveryPhrase.controller.js
```diff
@@ -2,105 +2,52 @@ angular
...
-    }, {
-      value: '',
-      actual: '',
-      index: null,
-      error: null
-    }
-  ];
+  $scope.recoveryPhrase = [];
+  $scope.words = [...Array(4)].map(_ => ({}));
+  $scope.addresses = Wallet.legacyAddresses().filter(a => a.active && !a.isWatchOnly && a.balance > 0);
```
Sjors: You may want to set the threshold slightly higher than 0, because of dust and transaction fees.

jtormey: So far I've been leaving that up to payment.js to decide (and showing a warning if there are funds that cannot be swept). What threshold do you think I should put here?

Sjors: You could add a method to `payment.js` to determine the threshold. But without a threshold I assume it will never stop nagging the user? Depending on where else we'll show this button/link in the future, that could be annoying.

PR #715, opened 2016-10-14
Missing localizations on landing page

Sjors:
We're sending these strings to our translator, but that can be done before merging. Please make sure I didn't break the layout before hitting merge.


Sjors:
Also I would suggest putting existing strings under `LANDING.` as well.

PR #742, opened 2016-10-28
SEPA Countries

plondon:


assets/js/controllers/coinify/coinifyCountry.controller.js
```diff
@@ -5,19 +5,179 @@ angular
 function CoinifyCountryController ($scope, country, MyWallet, buySell) {
   $scope.countries = country;
 
-  let blacklist = [{'Name': 'Algeria', 'Code': 'DZ'},
+  let blacklist = [{'Name': 'Albania', 'Code': 'AL'},
```
Sjors: It's probably more maintainable if we add a whitelist of SEPA countries, rather than an exhaustive list of non-SEPA countries. We could still keep the blacklist variable to prevent accidentally adding a country.

PR #763, opened 2016-11-08
Translations

Sjors:
Note that if you switch the landing page language to Chinese, the video will have Chinese audio.
Make sure I didn't break your video-playback directive....


Sjors:
@plondon Czech and Ukrainian are still hidden from the language dropdowns pending a backend change, but I fixed a few things so that they otherwise should work.

PR #768, opened 2016-11-09
fix(landingCtrl): get started with email input

prwelber:


app/landing.jade
```diff
@@ -148,7 +148,7 @@
     form.form-horizontal(
       role="form"
       name="signupForm"
-      ng-submit="signup()"
+      ng-submit="signup(fields.email)"
```
Sjors: This actually achieves nothing, because in the controller `$scope.signup = () =>` doesn't have any parameters. It gets `fields.email` directly from `$scope`.

assets/js/controllers/signup.controller.js
```diff
@@ -153,6 +153,8 @@ function SignupCtrl ($scope, $state, $cookies, $filter, $timeout, $translate, Wa
     email: $state.params.email || ''
   };
 
+  $state.params.email ? $scope.fields.emailIsFromState = true : '';
```
Sjors: Someone quickly reading this my see `emailIsFromState == true` instead of `=true` and not realize that this line is assigning a variable. The following would be better:
`$scope.fields.emailIsFromState = !!$state.params.email` or `$scope.fields.emailIsFromState = Boolean($state.params.email)`

PR #770, opened 2016-11-10
Get country options from server

Sjors:
This expects /Resources/wallet-options.json to exist. An example file is provided for local development.


Sjors:
This can go in the next prerelease.

PR #778, opened 2016-11-10
fix(home): imported address shows label

prwelber:


jtormey:
What if the address doesn't have a label?

PR #788, opened 2016-11-15
Bitcoin URL in Send

prwelber:


assets/js/controllers/send.controller.js
```diff
@@ -391,6 +391,30 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModal, $uibModalI
...
   };
 
+  $scope.handlePaste = (event) => {
+    let dest;
+
+    event.target.value
+      ? dest = event.target.value
+      : dest = $scope.transaction.destinations[0].address;
+
+    $timeout(() => {
```
Sjors: Can you add a comment why you need this timeout?

assets/js/controllers/send.controller.js
```diff
@@ -391,6 +391,30 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModal, $uibModalI
...
+      event.target.value
+        ? dest = event.target.value
+        : dest = $scope.transaction.destinations[0].address;
+
+      function extractFromUri (URI, RegEx, sliceAmount) {
```
Sjors: You may want to move `extractFromUri` into a service or helper, so that it's easier to write tests for it. I believe there is/was already a function somewhere that parses URL's (in that case from QR codes).

assets/js/controllers/send.controller.js
```diff
@@ -391,6 +391,30 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModal, $uibModalI
...
+        const data = URI.match(RegEx)[0];
+        return data.slice(sliceAmount, data.length);
+      }
+
+      $scope.transaction.destinations[0].address = extractFromUri(dest, /(?=\:)(.*)(?=\?)/, 1);
```
Sjors: In advanced send you may have multiple destinations. Is it still the intended behaviour to put the address in the first destination?

assets/js/controllers/send.controller.js
```diff
@@ -391,6 +391,30 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModal, $uibModalI
...
+      $scope.transaction.amounts[0] = extractFromUri(dest, /(?=amount=)(.*)(?=&message)/, 7) * 100000000;
+      $scope.transaction.note = decodeURI(extractFromUri(dest, /(?=&message=)(.*)/, 9));
```
Sjors: It might be better to process the URL in a helper method, which checks if it has the right format and extracts the address, amounts, and note. That helper method would return something like `{amount: 1, note: "Some note"}` to the controller. That keeps the controller simple.

PR #792, opened 2016-11-16
Buy page should use same rules as tab for when to show

Sjors:
For users with an account, but who's email address is not whitelisted (e.g. because they changed it) the production wallet currently has a bug. It will show the tab, but clicking on it won't work, because the routes file still uses the old rules.

This would also allow users in other countries to manually navigate to the buy tab, provided their email is whitelisted.

This fix still has duplicate code, so may need some refactoring.


Sjors:
I moved the code out of the router and into the buysell controller to make it slightly less likely for anyone to forget.

jtormey:
Nice

PR #794, opened 2016-11-17
Directive & Controller Inspector

Sjors:
Work in progress. The idea that you can go to `/parts/directives/adverts/` to inspect the `adverts` directive, etc, for all directives and controllers.

adverts

For each directive / controller, you can make an index.jade file which would contain minimum viable page. You can also mock its dependencies in mocks.js

In this example, the ad is hardcoded in the mock, so there's no network activity. 

At the moment most of the "magic" is still hard coded, so it only works for the adverts directive.


Sjors:
Superseeded by #1147

PR #796, opened 2016-11-21
Do not translate pointer strings

jtormey:
Checked the other locales as well, German was the only one affected.


Sjors:
That's quite likely to go wrong again. Can we avoid putting those strings in the locales file?

jtormey:
Well, these strings only work in the locales file. They tell angular-translate to use the translation for a different key. For example:

```js
{
  "CONTINUE": "Continue",
  "SEND_PRIMARY_BTN": "@:CONTINUE" // will resolve to "Continue"
}
```

We can make an effort to not use this convention though.

Sjors:
I see. I'll take that into account when integrating with the translation API service; maybe we can just skip sending those.

PR #800, opened 2016-11-29
Buy survey

plondon:


assets/js/controllers/coinify/coinify.controller.js
```diff
@@ -13,6 +13,10 @@ function CoinifyController ($scope, $filter, $q, MyWallet, Wallet, MyWalletHelpe
   $scope.trade = trade;
   $scope.quote = buyOptions.quote;
 
+  let links = ['https://blockchain.co1.qualtrics.com/SE/?SID=SV_8pupOEQPGkXx8Kp',
```
Sjors: I would still like to move these into `wallet-options.json` in a future release, to prevent the need for hot-fixes if these links change or a survey needs to be hidden.

PR #807, opened 2016-12-01
Fee stays below available balance

prwelber:


assets/js/controllers/send.controller.js
```diff
@@ -52,6 +52,9 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModal, $uibModalI
     tx.feeBounds = data.absoluteFeeBounds;
     tx.sweepFees = data.sweepFees;
     tx.size = data.txSize;
+    if (tx.fee > data.balance) {
```
Sjors: It might be better to do this in my-wallet-v3, cc @jtormey

PR #808, opened 2016-12-02
Onehour

Sjors:
Supersedes #797

Todo:
- [x] remove orphaned strings from locals files
- [x] upload namespaces (context)
- [x] upload English strings as of v1.8.3 (per namespace)
- [x] upload all human translations as of v1.8.3
- [x] upload newly created namespaces (done, pending API fix)
- [x] script to remove deleted strings
- [ ] script to upload new strings
- [ ] script to upload modified strings
- [ ] script to download most recent translations
- [ ] skip special strings such as `@:CONTINUE` (and write them back after download)
- [ ] script to check for common issues with translation (e.g. special characters)
- [ ] check if translation task is correct
- [ ] request translations
- [ ] await new translations
- [ ] add new translations to PR

Nice to have:
- [ ] check if any new strings were previously orphaned
- [ ] detect orphaned strings given a context (requires parsing an HTML subtree)
- [ ] use Jade files - or templates services, rather than minified Javascript
- [ ] extract into Ruby Gem


Sjors:
Let's try a different approach here.

PR #816, opened 2016-12-06
New buy-sell tab rules

Sjors:
Please double check my logic here...

* get sfoxWhitelist from config file
* always show buy-sell tab for Coinify countries
* move redirect to buySellSelectPartner controller
* buyStatus.canBuy() also returns true if user has an account


Sjors:
Oops, looks like I broke some tests. Not sure I get a chance to fix them before this release...

PR #823, opened 2016-12-08
SFOX Bug Fixes (Phil)

plondon:


app/partials/sfox/verify.jade
```diff
@@ -59,10 +59,7 @@
             i
           .item.width-33
             label State
-            ui-select(ng-model="state.state" name="state" required)
-              ui-select-match {{$select.selected.Name}}
-              ui-select-choices(repeat="s in states | filter:$select.search") {{s.Name}}
-              i
+            select(ng-model="state.state" name="state" ng-options="s as s['Name'] for s in states" tabindex="" required)
```
Sjors: Are you thinking of dropping `angular-ui-select` completely?

plondon: It would be nice, but <select> doesn't support search, which I think is a nice/useful feature.

PR #827, opened 2016-12-09
Format Zipcode

jtormey:


jtormey:
Commit message says "6 to 10 digit", actually should be "5 to 9" digit (code is correct though).

jtormey:
Rebased to avoid conflicts, and fixed commit message.

plondon:
If I enter zipcode `10021-1111` I get a 400 from SFOX with the following `{error: 'zipcode is required'}`

jtormey:
Possible that SFOX doesn't accept extended zip codes. Will clarify tomorrow.

plondon:
Also setAddress() saves the unaccepted zipcode to your profile, so even if you change zipcode to an acceptable format the request will fail again.

Should still be fixed on SFOX's end, or at least clarified like you said. Thanks!

jtormey:
Just going to use the standard 5-digit zip code.

assets/js/directives/format-input.directive.js
```diff
@@ -12,21 +12,26 @@ function formatInput () {
 
   function link (scope, elem, attrs, ctrl) {
     let inputFormat = attrs.formatInput;
-    let language = { digit: 'x' };
+    let language = { digit: 'x', open_group: '(', close_group: ')' };
```
plondon: Are open_group and close_group required still? Is this left over from zipcode 12345-(1211)?

jtormey: Not required, will remove.

jtormey: All set.

PR #859, opened 2016-12-19
Show invite form to percentage of users

Sjors:
- [x] `buyStatus.shouldShowInviteForm()`
- [x] actually show form 
  - [x] show tab if `canBuy || shouldShowInviteForm`
  - [x] show form if `shouldShowInviteForm`
- [x] SFOX should not appear in the dropdown if the user is outside the US (temporary)


Sjors:
One more thing that may still needs changing: SFOX should not appear in the dropdown if the user is outside the US

PR #860, opened 2016-12-19
Plaid Integration

plondon:
Depends on https://github.com/blockchain/bitcoin-sfox-client/pull/2

- [x] Add Tests
- [x] Finish account design (bank icons)
- [x] Don't blindly call arbitrary functions


Sjors:
If we pull Javascript from Plaid to render the widget, this has to happen inside an iframe.

So this needs to go:
`"script-src 'self' https://cdn.plaid.com/link/v2/stable/link-initialize.js"`

The easiest might be to add `child-src: self`. We would then create a plain HTML page that renders the widget and communicates the results to the parent (our wallet).

Similar to /landing.html, you would add /plaid.html. This requires a change to server.js and Grunt. This page needs to have its own Content Security Policy, so you want to duplicate `cspHeader` and make it as restrictive as we can.

plaid.html can be just a plain HTML file with some javascript, but we can also make it a simple AngularJS application. In fact, it may make sense to move it to a different repository and include it via NPM.

Sjors:
Here's a start, let me know if you get stuck:
schermafbeelding 2016-12-21 om 15 02 40

We'll have to change the CSP rules on staging/production for /wallet/plaid-*.html.

Sjors:
I'm replacing this with #883

PR #861, opened 2016-12-19
User friendly experience for 7 day trade restriction

jtormey:
Depends on
- [x] https://github.com/blockchain/bitcoin-coinify-client/pull/1
- [x] trade restriction added to / tested on Coinify staging

### DIFF
--- a/app/partials/buy-sell.jade
+++ b/app/partials/buy-sell.jade
@@ -13,7 +13,7 @@
                         modal-open="status.modalOpen"
                         change-currency="changeCurrency(currency)"
                         limits="limits"
-                        can-trade="exchange.profile.canTrade"
+                        trading-disabled="getIsTradingDisabled()"
                         buy="buy(null, {quote: quote, btc: btc, fiat: fiat, currency: transaction.currency})")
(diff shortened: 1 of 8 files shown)


Sjors:
It showed this error message when creating a new wallet:
new wallet

I was able to workaround that by adding a check for `exchange.hasAccount`.

However it seems that this error message always appears, even if Coinify doesn't set it. Not sure what to do about that...

app/partials/buy-sell.jade
```diff
@@ -13,6 +13,7 @@
                         modal-open="status.modalOpen"
                         change-currency="changeCurrency(currency)"
                         limits="limits"
+                        can-trade="exchange.profile.canTrade"
```
Sjors: `exchange.profile.canTrade` is undefined (falsy) when the user doesn't have an account or when the `canTrade` field is missing. A "negatrive" attribute would be safer here, e.g. `trade-suspended-until`.

jtormey: Updated to use a negative check (now `trading-disabled`)

jtormey:
Tested with 10-minute delay on staging. Works great.

PR #872, opened 2016-12-22
Add buy to Whats New

jtormey:
Will only show if the user can buy.

Before merging:
* [x] Get copy reviewed

### DIFF
--- a/app/templates/whats-new.jade
+++ b/app/templates/whats-new.jade
@@ -1,7 +1,6 @@
 li(
   ng-if="status.didUpgradeToHd"
   ng-repeat="feat in feats | orderBy:'-date' track by $index"
-  ng-hide="feat.title === 'BUY_BITCOIN' && !showBuyWhatsNew"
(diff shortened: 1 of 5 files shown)


app/templates/whats-new.jade
```diff
@@ -1,4 +1,12 @@
-li(ng-if="status.didUpgradeToHd" ng-class="::{flash: feat.date > lastViewedWhatsNew}" ng-repeat="feat in feats | orderBy:'-date' track by $index")
+li(
+  ng-if="status.didUpgradeToHd"
+  ng-repeat="feat in feats | orderBy:'-date' track by $index"
+  ng-hide="feat.title === 'BUY_BITCOIN' && !showBuyWhatsNew"
```
Sjors: Can you modify the list in the controller instead of the view?

How does the unread count behave?

jtormey: Thanks for catching that. Updated so that the filter is applied before the count is displayed.

PR #873, opened 2016-12-22
prevents main btc balance from breaking

prwelber:


prwelber:
Thanks for the feedback - will avoid in the future.

PR #877, opened 2016-12-27
Confirm trade in modal before initiating trade

plondon:


plondon:
I'll just restart this.

PR #878, opened 2016-12-28
recovery phrase before second pw

prwelber:


app/partials/settings/security.jade
```diff
@@ -46,7 +46,8 @@ h5.well.type-h5.em-400.hidden-xs(translate="SECURITY_BASIC_EXPLAIN")
         p.em-300.text.hidden-xs(translate="WALLET_SECOND_PASSWORD_EXPLAIN" ng-hide="userHasExchangeAcct")    
         p.em-300.text.hidden-xs(translate="WALLET_SECOND_PASSWORD_DISABLED_EXPLAIN" ng-show="userHasExchangeAcct")
       .col-sm-12.col-md-6.setting-result(ng-hide="active || removing")
-        button.button-primary.display-inline-block(ng-click="activate()", translate="SET_SECOND_PASSWORD", ng-show="!settings.secondPassword" ng-disabled="userHasExchangeAcct")
+        button.button-primary.display-inline-block(ng-click="activate()", translate="SET_SECOND_PASSWORD", ng-show="showButton()" ng-hide="hideButton()" ng-disabled="userHasExchangeAcct")
+        button.button-primary.display-inline-block(ng-click="promptRecovery()", translate="SET_SECOND_PASSWORD", ng-hide="walletStatus.didConfirmRecoveryPhrase || settings.secondPassword || walletStatus.dismissedRecoveryPrompt" ng-show="!settings.secondPassword && !walletStatus.didConfirmRecoveryPhrase" ng-disabled="userHasExchangeAcct")
```
jtormey: Can these be the same element? You might be able to do something like `ng-click="walletStatus.didConfirmRecoveryPhrase ? activate() : promptRecovery()`. That would simplify the show / hide logic as well.

assets/js/controllers/settings/manageSecondPassword.controller.js
```diff
@@ -61,4 +69,44 @@ function ManageSecondPasswordCtrl ($rootScope, $scope, Wallet, $timeout, MyWalle
...
+
+  $scope.recoveryModal = () => {
+    const openModal = () => $uibModal.open({
+      templateUrl: 'partials/recovery-before-second-password.jade',
+      controller: 'RecoveryBeforeSecondPasswordCtrl',
+      windowClass: 'bc-modal'
+    });
+
+    if (!Wallet.status.didConfirmRecoveryPhrase && !Wallet.settings.secondPassword) {
+      openModal();
```
jtormey: `$uibModal.open(...).result` is a promise that rejects if the modal was "dismissed" and resolves if the modal was "closed". Using this promise to detect what the user does with the modal would be more clean than having a `dismissedRecoveryPrompt` variable and watching it. Let me know if you want to talk about this, would be happy to help.

assets/js/controllers/settings/manageSecondPassword.controller.js
```diff
@@ -61,4 +69,44 @@ function ManageSecondPasswordCtrl ($rootScope, $scope, Wallet, $timeout, MyWalle
...
+    if (confirmed && !secondPW) {
+      return false;
+    } else if (secondPW) {
+      return true;
+    } else if (!dismissed || !confirmed) {
+      return true;
+    } else if (dismissed && secondPW) {
+      return true;
+    }
+  };
```
jtormey: I mentioned this in a previous comment, but I think this show / hide logic could be simplified a lot by having a single button, and really considering what it's purpose is.

assets/js/directives/settings-form.directive.js
```diff
@@ -12,12 +12,20 @@ function settingsForm () {
       scope.$parent.active = false;
       scope.$parent.reset();
 
+      scope.$parent.promptRecovery = () => {
+        scope.$parent.promptForRecovery = true;
+        scope.$parent.recoveryModal();
+      };
```
jtormey: This directive is meant to be an abstraction for every setting form. Could you leave this functionality in the manageSecondPassword controller?

PR #887, opened 2017-01-05
Plaid Webpack

jtormey:
This was a good exercise, it nicely shows how a webpack-based angular app would look. If we like how it feels (definitely a lot less configuration so far) we might want to start transitioning pieces of the wallet app over as well.


jtormey:
For those interested, check out the guide I took inspiration from: http://angular-tips.com/blog/2015/06/using-angular-1-dot-x-with-es6-and-webpack/

Sjors:
Looks like something went wrong with grunt dist despite your last commit:
schermafbeelding 2017-01-06 om 09 42 51

Might be a sequence thing.

Sjors:
I'm not a huge fan of `grunt-exec`. More generally I'm trying to move things out of this huge `Gruntfile` and into the `Makefile`, especially if it's just executing a shell command.

helperApp/plaid/plaid.scss
```diff
@@ -1 +1,10 @@
-@import "../assets/css/blockchain.css.scss";
+
+@import "../../bower_components/bootstrap-sass/assets/stylesheets/bootstrap/variables";
+@import "../../bower_components/bootstrap-sass/assets/stylesheets/bootstrap/mixins";
+@import "../../bower_components/bootstrap-sass/assets/stylesheets/bootstrap/buttons";
+@import "../../assets/css/utilities/colors";
+@import "../../assets/css/elements/buttons";
+
```
Sjors: Nice, I wasn't able to figure out what the minimum required set of CSS was.

helperApp/plaid/index.html
```diff
@@ -1,21 +1,12 @@
...
-  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.10/angular-sanitize.min.js" defer></script>
-  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.10/angular-route.min.js" defer></script>
-  <!-- @if !PRODUCTION -->
-  <script src="plaid.js" defer></script>
-  <script src="plaid.controller.js" defer></script>
-  <!-- @endif -->
-  <!-- @if PRODUCTION !>
-  <script src="plaid.min.js" defer></script>
-  <!-- @endif -->
+  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.10/angular.min.js"></script>
```
Sjors: Did you mean to remove `defer`? (I'm not sure if it was needed here)

Gruntfile.coffee
```diff
@@ -579,10 +589,12 @@ module.exports = (grunt) ->
   grunt.loadNpmTasks('grunt-karma-coveralls')
   grunt.loadNpmTasks('grunt-text-replace')
   grunt.loadNpmTasks('grunt-include-source')
+  grunt.loadNpmTasks('grunt-exec')
 
   grunt.registerTask "build", [
     "html2js"
     "babel:build"
+    "exec:webpack"
```
Sjors: I would just do this as part of `make build`

Sjors:
I think you were using a more modern version of NodeJS on your dev machine. There was some ES6 magic in `webpack.config.js`. Would be nice to bump the NodeJS version of the frontend if there are no side-effects.

helperApp/plaid/plaid.js
```diff
@@ -1,25 +1,16 @@
+require('./plaid.scss');
```
Sjors: I first thought that this would include the CSS inside the Javascript, but it nicely generates `plaid.css`. Such magic!

Sjors:
Ok, I made a few minor tweaks and it's good to merge now. Looks really nice, much more compact.

helperApp/plaid/plaid.scss
jtormey: I did that mostly because I was having trouble configuring webpack to handle fonts. Will have to cross that bridge eventually though.

helperApp/plaid/index.html
jtormey: I removed it because the app was loading before angular, would throw an "angular is not defined error".

helperApp/plaid/plaid.js
jtormey: We can include the css inside the Javascript, webpack does that by default. However it was causing CSP issues, we'd have to tweak those if we want everything inside the js bundle. That might be no bueno for security though.

Sjors: Separate file is better in my view.

PR #888, opened 2017-01-06
sfox column header alignment

prwelber:


prwelber:
@plondon This introduces a new css class, mainly because we did not have existing classes for the text-indent property, which I found to work best here in terms of having the status column data aligned with the same starting point, but not all the way to the edge of the TD/TH. Let me know if you think it needs another refactor.

PR #898, opened 2017-01-11
Add checkout process to last step of signup

jtormey:
A lot of this was just moving code around, so don't be too freaked out by the huge diffs.


app/templates/buy-checkout.jade
```diff
@@ -0,0 +1,72 @@
...
+          span
+            div.flex-row.flex-center(ng-show="state.rate && !state.loadFailed")
+              span 1 {{ bitcoin.code }} = ${{ state.rate }}
+              helper-button(content="EXCHANGE_RATE_HELPER")
+            .state-danger-text(ng-show="state.loadFailed")
+              i.ti-alert
+              span.mls Failed to load exchange data
+        .flex-center
+          .flex-1.flex-center.border
+            input.form-control.no-border(
```
plondon: Do you think it would make sense to add a focus-when here? Might be tricky because it will effect both sfox/checkout and onStep('buy')

jtormey: Might be nice to have that input auto-focus in both places, will try it out.

jtormey: The autofocusing is nice, added it.

PR #904, opened 2017-01-16
Record metrics for address input method

jtormey:


assets/js/controllers/send.controller.js
```diff
@@ -179,6 +186,11 @@ function SendCtrl ($scope, $log, Wallet, Alerts, currency, $uibModal, $uibModalI
       .then(transactionSucceeded).catch(transactionFailed);
   };
 
+  $scope.sendInputMetrics = (metric) => {
+    let root = $rootScope.rootURL ? $rootScope.rootURL : '/';
+    $http.get(`${root}event?name=wallet_web_tx_from_${metric}`);
```
Sjors: Is this the same analytics endpoint as we use to track 2nd password usage? If so, can you move this to My-Wallet-V3?

jtormey: It is the same endpoint. I don't know if I would consider this the responsibility of My-Wallet-V3 though, since the events are specific to the web wallet.

Sjors: The events are, but My-Wallet-V3 could have a method like `trackEvent(name)`.

PR #905, opened 2017-01-16
New ads system

alecalve:


assets/js/services/adverts.service.js
```diff
@@ -21,16 +21,27 @@ function Adverts ($http, $rootScope) {
   }
 
   function fetch () {
-    let advertsFeed = $rootScope.rootURL + 'adverts_feed?wallet_version=3';
+    if (!$rootScope.apiDomain.endsWith('blockchain.info/')) {
```
Sjors: Dev and staging need to have ads too, otherwise we can't test this for layout and bug issues. So I think this check can go away.

assets/js/services/adverts.service.js
```diff
@@ -21,16 +21,27 @@ function Adverts ($http, $rootScope) {
...
     $http.get(advertsFeed)
-      .success(data => {
-        let adverts = data.partners.home_buttons.splice(0);
-        service.ads.push(randFromArray(adverts));
-        service.ads.push(randFromArray(adverts));
-      });
-  }
+      .success(function (data) {
+        data.forEach(function (ad) {
+          var a = document.createElement('a');
```
Sjors: This is too jQuery-ish. @jtormey and @plondon should be able to help you do this the Angular Way. It would probably involve an `ng-repeat` in the template.

assets/js/services/adverts.service.js
```diff
@@ -21,16 +21,27 @@ function Adverts ($http, $rootScope) {
...
+          a.href = ad.url;
```
Sjors: Let's not hard-code a URL into the ad feed; it's too tedious and risky to parse it. Instead, just construct the URL in the ad rendering code

So the JSON have `"id": "1234"`

And then the explorer / wallet would send the user to `$rootScope.apiDomain + "/ads/" + ad.id`

Sjors: You should still check `ad.id` with a regex. I suggest making the id an integer field, so you can just check the type.

assets/js/services/adverts.service.js
```diff
@@ -21,16 +21,27 @@ function Adverts ($http, $rootScope) {
...
+
+          if (!a.protocol.startsWith('http') ||
+              a.host === '' ||
+              a.host === window.location.host ||
+              !/^data:image\/(png|jpg|jpeg|gif);base64,/.test(ad.data) ||
```
Sjors: with the above I don't think you need any of these tests

assets/js/directives/adverts.directive.js
```diff
@@ -20,7 +20,8 @@ function adverts (Adverts, $window) {
     Adverts.fetchOnce();
 
     scope.visit = (ad) => {
-      $window.open(ad.link, '_blank');
+      var url = $rootScope.apiDomain + 'ads/out?id=' + ad.id;
+      $window.open(url, '_blank');
```
Sjors: Can you use `ng-href` in the view instead of `$window.open()` and then add `noopener noreferrer`?

assets/js/services/adverts.service.js
```diff
@@ -21,16 +21,22 @@ function Adverts ($http, $rootScope) {
...
-        let adverts = data.partners.home_buttons.splice(0);
-        service.ads.push(randFromArray(adverts));
-        service.ads.push(randFromArray(adverts));
-      });
-  }
+      .success(function (data) {
+        data.forEach(function (ad) {
+          if (!/^data:image\/(png|jpg|jpeg|gif);base64,/.test(ad.data) ||
+              !angular.isNumber(ad.id) ||
+              !/^[0-9a-zA-Z ]*$/.test(ad.name)) {
```
Sjors: This might be too strict for non-latin alphabet advertisers, but can expand that later if it's not an issue yet.

alecalve: My approach would be to make it restrictive at first and then accepting non latin chars if wanted by partners. So far the existing ads are not affected by it and it's impossible to create some that would not pass this check.

PR #909, opened 2017-01-18
Buy history table alignment

prwelber:


prwelber:
less declarative, but makes sense. removed classes.

PR #919, opened 2017-01-19
Sfox fixes

plondon:
Code additions are mostly tests


assets/js/components/buy-checkout.component.js
```diff
@@ -95,6 +93,12 @@ function BuyCheckoutController ($scope, $timeout, $q, currency, Wallet, MyWallet
     this.handleBuy({ quote }).finally($scope.resetFields).finally($scope.free);
   };
 
+  $scope.setLimits = (limit) => {
+    $scope.min = currency.convertToSatoshi(0.01, $scope.dollars);
+    $scope.max = currency.convertToSatoshi(limit, $scope.dollars);
+  };
```
plondon: Can change this when SFOX also sends min limit.

assets/js/components/buy-checkout.component.js
```diff
@@ -95,6 +93,12 @@ function BuyCheckoutController ($scope, $timeout, $q, currency, Wallet, MyWallet
...
+
+  $scope.$watch('$ctrl.buyLimit', (limit) => limit && $scope.setLimits(limit));
```
plondon: I think found an interesting bug here, when the limit is 0 this won't set the limits. Will fix.

PR #920, opened 2017-01-20
Quality

jtormey:


assets/js/services/modals.service.js
```diff
@@ -9,6 +9,15 @@ function modals ($state, $uibModal, $ocLazyLoad) {
...
 
+  service.openOnce = (modalOpener) => {
+    let modalInstance = null;
+    return (...args) => {
+      if (modalInstance) return;
+      modalInstance = modalOpener(...args);
+      modalInstance.result.finally(() => { modalInstance = null; });
+    };
+  };
+
```
plondon: This is great, can we use it on SFOX too?

jtormey: Yeah, good idea. Can be used on any function that returns a `$modalInstance`.

PR #921, opened 2017-01-20
Tx feed fixes

prwelber:


plondon:
Can we move this to 1.11 since https://github.com/blockchain/My-Wallet-V3-Frontend/pull/925 is effected?

prwelber:
Sounds good. 

Also- we're not top aligning all the tx feed elements, since they have 1, 2 or 3 lines of text. This fix should center elements vertically instead.

PR #923, opened 2017-01-23
Faster Builds

plondon:
Only calls shell:webpack when helperApp changes are made. Brings template build down from 20-22sec to 5-6sec.


prwelber:
bless you @plondon

Sjors:
I'll fix it on `master` by adding `shell:webpack` back into build, but making sure none of the watchers call `build`.

Sjors:
It should also be possible to make Webpack faster. I think it's scanning all the `node_modules` folders or something slow like that.

PR #938, opened 2017-01-30
Tooltip style change

prwelber:


plondon:
Are these supposed to match .popover-content?

tooltip
popover-content

Tooltips seem to be a bit short.

Also is there a way we can just `@extend .popover-content` in anyway?

PR #945, opened 2017-02-01
Simplify wallet-options

Sjors:


plondon:
Will we get rid of wallet-options-debug.json in info-resources as well?

Sjors:
This has been superseeded.

PR #956, opened 2017-02-07
Center amount

prwelber:


plondon:
Nice, I also think the font should be Montserrat for these?

I also think Sent, Received, Transferred is Montserrat and uppercase.

PR #962, opened 2017-02-07
Handle wrong 2FA through promise rejection

jtormey:
- [x] https://github.com/blockchain/My-Wallet-V3/pull/327


Sjors:
I may have screwed up the merge, but my 2FA field is gray after the first wrong attempt:

2fa wrong gray

jtormey:
@Sjors, I found the issue is that the Google Auth code error messages are different from SMS code error messages (I only tested with the latter). This was causing `errorCallback` to fire without a `field` argument passed in. Fixed now, although I'd like your thoughts on coming up with a way to do this without relying on testing the message for strings using `indexOf`... seems fragile.

PR #969, opened 2017-02-08
Send full image url to modal helper

plondon:


assets/js/services/modals.service.js
```diff
@@ -27,7 +27,10 @@ function modals ($state, $uibModal, $ocLazyLoad) {
   };
 
   service.openHelper = (helper) => open({
-    controller ($scope) { $scope.helper = helper; },
+    controller ($scope) {
+      $scope.helper = helper;
+      $scope.image = 'img/' + helper + '.png';
```
Sjors: This won't solve the issue of `grunt dist` not renaming the files.

Sjors: Try something like:
```
let helperImages = {
  'bank' : 'img/bank.png',
  'something': 'img/something.png'
}
$scope.image = helperImages[helper];
```

PR #974, opened 2017-02-10
No more #

Sjors:


Sjors:
This replaces #98.

Moving this to a later release.

Sjors:
Rebased, unfortunately `base(href='/wallet/')` breaks our new language switching convention (e.g. `/de/wallet`). This begs the question why the wallet, unlike the explorer, is using a URL to set the language rather than the `clang` cookie. cc @mpfluger

mpfluger:
Wallet, explorer and charts all have the same logic for determining the language (in order):

req path (/de/wallet)
req parameter language (?language=de)
clang cookie
locale

Only the language parameter changes the cookie.

Sjors:
In that case it might make sense for /wallet/lang to change the language cookie and then redirect to /wallet. That would mean this logic is handled by angular, not server side.

Sjors:
Closing due to secret alternative plan :-P

PR #995, opened 2017-02-23
Extract Global CSS

plondon:


assets/css/blockchain.css.scss
```diff
@@ -1,31 +1,42 @@
...
+
+//import global elements
+@import '../../node_modules/blockchain-css/sass/elements/buttons';
+@import '../../node_modules/blockchain-css/sass/elements/inputs';
+@import '../../node_modules/blockchain-css/sass/elements/forms';
+@import '../../node_modules/blockchain-css/sass/elements/links';
+
+//import global modules
+@import '../../node_modules/blockchain-css/sass/modules/header';
+@import '../../node_modules/blockchain-css/sass/modules/footer';
```
jtormey: Can you drop the node_modules part, so that it's just `@import blockchain-css/...`? If it doesn't work by default, pretty sure you can add something to the Gruntfile to include that dir when sass is looking for modules.

plondon: Good idea

PR #997, opened 2017-02-24
Jade becomes Pug

Sjors:
Jade has been renamed to Pug a while ago due to some trademark issue.

For the most part I just had to rename all the files. However there are two dependencies not updated yet, so I'm using a fork for those. We may not need those anyone if we move to webpack.

Pug also comes with a linter, so grunt build will fail if you e.g. use incorrect indentation. I had to fix that in two places.

There's a bunch of references like `include="'partials/sfox/verify.pug'`. I didn't touch those, but I think we should configure grunt such that it renames the partials to .html first, since the specific template language shouldn't matter when including them.

cc @prwelber


Sjors:
jwvdiermen/grunt-include-source#52 was just merged, so I won't need to use a fork of that.

plondon:
🐶

PR #1000, opened 2017-02-28
Backup Plan for Third Party Services

plondon:
The exchange partner will no longer drop out of service completely if a call to /trades or /kycs fails upon initialization. Instead, a specific error message will show for each.

If the metadata service is down, a banner error will show (EXCHANGE_DOWN).
If logging into your exchange partner fails, a different banner error will show (METADATA_DOWN).

Trading can be disabled at anytime by flipping the switch on wallet-options -> partners -> partner -> disabled -> true. https://github.com/blockchain/My-Wallet-V3-Frontend/pull/1000/files#diff-e071ef8988ca957e1d212979614c33deR5

For traders that already have trades, trade history and kyc will become read only features.


jtormey:
Awesome

PR #1008, opened 2017-03-01
Some old browsers ignore 'disabled'

Sjors:


app/partials/login.jade
```diff
@@ -56,5 +56,5 @@
   .ptl.flex-end
     div(ng-hide="didLogout")
       span(translate="HAVING_TROUBLE")
-      |
```
jtormey: I think this space is intentional.

app/partials/signup.jade
```diff
@@ -5,15 +5,15 @@ div(data-preflight-tag="CreateWallet")
         span.f-24.lh1(translate="CREATE_A_WALLET")
         span
           span(translate="OR")
-          |
```
jtormey: This space seems like it also might have been intentional.

Sjors:
Making a new PR. I'll check the spacing.

PR #1021, opened 2017-03-07
Pull n users from charts data

plondon:
@alecalve What's the status of my-wallet-n-users on dev/staging?

### DIFF
--- a/assets/js/landingCtrl.js
+++ b/assets/js/landingCtrl.js
@@ -1,6 +1,6 @@
 angular.module('walletApp').controller('LandingCtrl', LandingCtrl);
 
-function LandingCtrl ($scope, $state, $sce, languages) {
+function LandingCtrl ($scope, $state, $sce, $http, languages) {
   $scope.fields = {
     email: undefined
   };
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -54,6 +54,7 @@ app.use(function (req, res, next) {
         rootURL,
         (webSocketURL || 'wss://ws.blockchain.info'),
         (apiDomain || 'https://api.blockchain.info'),
+        'https://api.blockchain.info',
```
plondon: Apparently there is no way to get the data from dev, and staging is down. Will it cause problems to have the same apiDomain when on production? @Sjors

Sjors: You shouldn't have a duplicate entry. In `server.js` you can prevent that with e.g. `apiDomain === `https://api.blockchain.info` ? '' : 'https://api.blockchain.info'`

`server.js` is only used for local development, not for deployment, so headers need to modified for that too.

assets/js/landingCtrl.js
```diff
@@ -35,4 +35,11 @@ function LandingCtrl ($scope, $state, $sce, languages) {
     $scope.language = languages.mapCodeToName(code);
     $scope.searchUrl = code === 'en' ? '/search' : `/${code}/search`;
   });
+
+  let setNWallets = (res) => {
+    let nWallets = res.data.values[res.data.values.length - 1]['y'];
+    $scope.nWallets = Math.floor(nWallets / 1000000);
+  };
+
+  $http.get('https://api.blockchain.info/charts/my-wallet-n-users?format=json').then(setNWallets);
```
Sjors: Don't forget to handle a failed request, a missing and a malformed response. It should probably fall back to some safe value in that case.

plondon: It just won't show the nWallet number and the UI will fallback gracefully. We could hardcode a "safe value", but that kind of defeats the purpose of this.

plondon:
Closing in favor of more recent implementation that includes fallback

PR #1022, opened 2017-03-13
Payment request

Sjors:


assets/js/controllers/settings/addresses.controller.js
```diff
@@ -2,75 +2,86 @@ angular
...
+  $scope.totalPast = $scope.addresses.length - totalLabeled;
 
   $scope.createAddress = () => {
-    Wallet.addAddressForAccount($scope.account)
-      .then(address => $scope.paymentRequests.push(address))
-      .catch(Alerts.displayError);
+    $scope.createAddressInProgress = true;
+
+    Labels.addLabel(accountIndex, 15, $translate.instant('DEFAULT_NEW_ADDRESS_LABEL'))
+      .then(address => $scope.addresses.push(address))
```
plondon: This is causing some issues and I don't think we need it.

Sjors: I'll double check that, but if I remember correctly, `Labels.all()` returns a clone of the array of addresses, not the original one. So adding an address, wouldn't update the cloned copy. What problem are you seeing?

plondon: `$scope.createAddress()` creating the same address.

assets/js/controllers/settings/addresses.controller.js
```diff
@@ -2,75 +2,86 @@ angular
...
+  let accountIndex = parseInt($stateParams.account, 10);
+
+  $scope.presentFilter = (address) => address && !!address.label && address.used === false;
+
+  $scope.pastAddressesPage = (page, pageLength) => {
+    let pastFilter = (address) => address && (!address.label || address.used === true);
+
+    return $scope.addresses.filter(pastFilter)
+                           .reverse().slice((page - 1) * pageLength, page * pageLength);
+  };
```
plondon: Awesome refactor. I noticed on production though that the pastAddresses section is initiated with the first address in the account with (No Transactions), but in this implementation it's empty.

plondon: Also, following these steps from @jtormey to create a "used" address I get an error:

Steps: Label a new address -> Send to current receive address / Label another address -> Remove label from address in step 1

Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: address in addresses | filter:presentFilter | filter:searchText | orderBy:'-', Duplicate key: object:1227, Duplicate value: {"label":"New Address"}

Sjors: I also just noticed the duplication issue. Will fix that. Hopefully delete is the same issue...

I'll look into the issue of the first available address in the past addresses section.

Sjors: If fixed the duplication issue.

PR #1028, opened 2017-03-14
Warning to discourage sharing of shared key

jtormey:


kristovatlas:
Good idea to warn about this. 👍

PR #1063, opened 2017-04-05
Parse full bitcoin uri

plondon:


plondon:
Discussed with Justin, this is an error in the way I'm generating the URL on the explorer side.

PR #1068, opened 2017-04-06
Second password / Exchange account

jtormey:
@Sjors can you confirm that this is correct? Specifically, that exchange account is not dependent on second password?


Sjors:
If a user has a second password and goes to the buy-sell tab, two things can happen:
1 - if wallet.metadataNode is set, it works
2 - if it's not set, the user is prompted for their second password; if they don't provide it, user is sent away from that tab

Setting a second password is no longer a problem for buy sell, as long as they enter it when need to set wallet.metadataNode. cc @pernas

jtormey:
Ok, so sounds like this is the correct fix then.

PR #1071, opened 2017-04-07
Buy Refactor

plondon:
- [x] Handle quote expiration
- [x] Handle new quote after account creation
- [x] Load completed/incomplete trades
- [x] Fix tests and improve test coverage
- [x] installLock
- [x] Improve summary controller and editAmount
- [x] Limit handling
- [x] Bank transfers || KYC logic
- [x] Surveys/Close/Back Button
- [ ] Improve formatTrade step and ISX communication with wallet
- [x] Don't break sell w/ @prwelber 
- [x] Don't break mobile buy w/ @jtormey


plondon:
Depends on https://github.com/blockchain/bitcoin-exchange-client/pull/3

plondon:
For limits, still waiting on updated API spec from Coinify before handling this better.

app/partials/coinify-modal.pug
```diff
@@ -1,106 +1,60 @@
...
-          ng-show="onStep('summary') && !error")
-      div(ng-show="onStep('trade-in-review') || onStep('summary') && error || onStep('trade-formatted') && !bitcoinReceived")
-        button.button-primary(translate="CLOSE" ng-click="cancel() && goToOrderHistory()")
-      div(ng-show="onStep('trade-formatted') && bitcoinReceived")
-        button.button-primary(translate="SEE_BITCOIN" ng-hide="inMobileBuy" ng-click="cancel()" ui-sref="wallet.common.transactions")
-        button.button-primary(translate="SEE_BITCOIN" ng-show="inMobileBuy" ng-click="exitToNativeTx()")
+  
+  div(ng-if="onStep('email')" ng-controller="CoinifyEmailController" ng-include="'partials/coinify/email.pug'")
+  div(ng-if="onStep('accept-terms')" ng-controller="CoinifySignupController" ng-include="'partials/coinify/signup.pug'")
+  div(ng-if="onStep('select-payment-medium')" ng-controller="CoinifyMediumController" ng-include="'partials/coinify/medium.pug'")
```
Sjors: It would be nice if you can move these `ng-include`s into their own components.

assets/js/controllers/coinify/coinify.controller.js
```diff
@@ -2,259 +2,132 @@ angular
...
     });
   };
 
   $scope.close = () => {
     let index;
+    let links = options.partners.coinify.surveyLinks;
 
-    if (!$scope.exchange.user) index = 0;
-    else if (!$scope.trades.length && !$scope.trade) index = 1;
+    if (!this.exchange.user) index = 0;
```
Sjors: Nice!

plondon:
@sjors, would love it if we moved limits to bitcoin-coinify-client so Frontend can do `limit.min[curr] = exchange.min[curr]`

Sjors:
@plondon that sounds good

app/partials/coinify-modal.pug
```diff
@@ -1,106 +1,60 @@
...
+  div(ng-if="onStep('summary')" ng-controller="CoinifySummaryController" ng-include="'partials/coinify/summary.pug'")
```
Sjors: If you move each step into its own component / directive, you could do something like this:
```
coinify-summary(on-complete='nextStep()', ng-if="onStep('summary')")
```
This CoinifyController would be responsible for figuring out the current step and which step is next. Each component should only care about check if it's step is complete and then call `scope.onComplete()`.

You can take it one step further with ui-router, which works inside modals:
https://github.com/angular-ui/ui-router/wiki/Frequently-Asked-Questions#how-to-open-a-dialogmodal-at-a-certain-state

Each step is then a route, e.g. `coinify.summary`, so you don't need `onStep()`. Instead, the CoinifyController is the parent state `coinify` and it takes care of sending the user to the next step (as well as sending the user back to a previous step if they try to nagivate to a future step too early).

Since both the individual step component and the CoinifyController need to know the criteria for when a step is completed, I think CoinifyController should have keep an array of function to track this. E.g.

```js
scope.isCompleted = [(a, b) => a & b]
```

Although at that point it might be better to use a service, and perhaps the routes file is a more appropriate place to handle this. @jtormey?

plondon: I like this idea, the reason I chose to use controllers was to have easy access to `$scope.vm.quote` or `$scope.vm.exchange`. In `accept-terms` for example, the parent scope needs to be refreshed before continuing, otherwise you get to summary with a fake quote. I'd rather set an updated property to `$scope.vm.property = property` than using directives, which always get messy. What is the downside of doing it this way?

plondon: We can still do take advantage of using ui-router either way.

Sjors: The downside is entanglement. It's hard to test these parts in isolation if they can mess with scope outside of themselves. Passing `quote` and `exchange` as attributes into a directive would be better I think.

jtormey: Passing state from attrs / callbacks is good because it creates a very clear boundary between the component and the parent controller (although `vm` is usually fine too, as long as children treat it as read-only like we try to do with sfox).

Would be cool to use ui-router. There are some issues I can see with that though. If we start opening the modal with a route change instead of the `modal` service, can we still pass in data like a `trade` object? Would that be done through state params? Also, would we lose the benefit of receiving a promise from opening the modal?

Sjors: Good questions. Passing a trade could be achieve by adding the trade id to the state params (and then looking up the trade again through some service). But that might not work well with quotes, since these expire and get refreshed all the time. I'm not sure if you can swap out id's all the time without closing and reopening the modal. If possible though, it would cause a much stronger seperation of scope; it's not ideal to pass these complex objects (with side-effects) around.

Sjors: Worst case the current quote could part of the global application state, which is not much worse than the current situation.

Sjors: (I would argue it's better, since it's more explicit that this thing represents global state that we need to deal with)

Sjors: One more thought, regarding quotes, the URL parameters could be the amount. E.g. `/buy/coinify/10/eur` and then the controller / component will reuse an existing quote or create a new one based on that.

Sjors: That would also make calls to action like "would you like to buy €25 more today?" very easy to implement.

jtormey: Would rather not pollute the url bar too heavily. Could use state params for some of those pieces of data to the same effect. But yeah, nice idea.

assets/js/controllers/buySell.controller.js
```diff
@@ -59,16 +58,10 @@ function BuySellCtrl ($rootScope, $scope, $state, Alerts, Wallet, currency, buyS
...
-      let curr = $scope.transaction.currency || null;
-      $scope.currencySymbol = currency.conversions[curr.code];
-      if (newVal !== oldVal) $scope.getMaxMin($scope.limits, $scope.transaction);
-    });
-
     $scope.$watch('sellTransaction.currency', (newVal, oldVal) => {
       let curr = $scope.sellTransaction.currency || null;
       $scope.sellCurrencySymbol = currency.conversions[curr.code];
-      if (newVal !== oldVal) $scope.getMaxMin($scope.sellLimits, $scope.sellTransaction);
+      // if (newVal !== oldVal) $scope.getMaxMin($scope.sellLimits, $scope.sellTransaction);
```
plondon: @prwelber we'll need to figure out how to handle max min amounts for sell. Maybe you can do something similar to what I'm doing for now with buy.

assets/js/services/buySell.service.js
```diff
@@ -153,24 +155,36 @@ function buySell ($rootScope, $timeout, $q, $state, $uibModal, $uibModalStack, W
...
+      service.limits.card.max[curr] = service.calculateMax(rate, 'card');
+      service.limits.absoluteMax = (curr) => service.limits.bank.max[curr] > service.limits.card.max[curr] ? service.limits.bank.max[curr] : service.limits.card.max[curr];
+    };
+
+    let DKK = service.getRate(defaultCurrency, 'DKK').then((rate) => calculateMax(rate, 'DKK'));
+    let EUR = service.getRate(defaultCurrency, 'EUR').then((rate) => calculateMax(rate, 'EUR'));
+    let USD = service.getRate(defaultCurrency, 'USD').then((rate) => calculateMax(rate, 'USD'));
+    let GBP = service.getRate(defaultCurrency, 'GBP').then((rate) => calculateMax(rate, 'GBP'));
+
+    return $q.all([DKK, EUR, USD, GBP]);
```
plondon: @jtormey know a nicer way to do this?

jtormey:
```
let getRate = (c) => service.getRate(defaultCurrency, c).then(r => calculateMax(r, c))
return $q.all(['DKK', 'EUR', 'USD', 'GBP'].map(getRate))
```
Thoughts?

jtormey:
@plondon yup ng-if works too. The issue was that the templates would render before angular had time to apply the ng-hide class, causing a flicker. With ng-if, they don't render in the first place.

plondon:
Buy and Sell refactors were combined here: https://github.com/blockchain/My-Wallet-V3-Frontend/pull/1117. Closing this.

PR #1077, opened 2017-04-12
Coinify-email -> Component

jtormey:


assets/js/components/coinify/email.component.js
```diff
@@ -0,0 +1,32 @@
+angular
+  .module('walletApp')
+  .component('coinifyEmail', {
+    bindings: {
+      email: '<',
```
plondon: Do we need to pass in email here? Should the email component set this itself?

assets/js/controllers/coinify/coinify.controller.js
```diff
@@ -10,7 +10,7 @@ function CoinifyController ($scope, $filter, $q, MyWallet, Wallet, MyWalletHelpe
   $scope.alerts = [];
   $scope.status = {};
 
-  this.user = Wallet.user;
+  this.user = $scope.user = Wallet.user;
```
plondon: Necessary?

app/partials/coinify-modal.pug
```diff
@@ -23,8 +23,8 @@
           quote-countdown.mid-grey(ng-hide="isKYC" quote="quote" trade-obj="trade" expired-quote="expiredQuote")
           p.security-red.pointer(ng-show="buySellDebug && quote.id")
             i.ti-search.mrm.right-align(ng-click="quote.expire()")  QA: Now
-  
-  div(ng-if="onStep('email')" ng-controller="CoinifyEmailController" ng-include="'partials/coinify/email.pug'")
+
+  coinify-email(ng-if="onStep('email')" on-complete="goTo('accept-terms')" email="user.email" verified="user.isEmailVerified" rejected-email="rejectedEmail")
```
plondon: Nice, maybe one day this will be `buy-email`

assets/js/controllers/coinify/coinify.controller.js
jtormey: Nope, fixed.

assets/js/components/coinify/email.component.js
jtormey: We don't need to, but I think it's a nicer way of dealing with data.

If we were to do `this.email = Wallet.user.email` when the component is initialized, now `email` is duplicate state. When we update `user.email` we have to remember to update `this.email` as well. If `email` is a binding, that is done automatically for us, and we have more control (if we ever want it) through `$onChanges`.

Could also do `this.user = Wallet.user` when the component is initialized, then `ng-model="$ctrl.user.email"` in the template. But, angular two-way binds by default, so I think this would end up changing `user.email` before we want it to. Also, exposing all of `Wallet.user` to the component seems unnecessary.

What do you think? Is there another way that would be cleaner / easier?

PR #1078, opened 2017-04-13
Sell Fixes

prwelber:


assets/js/components/sell-quick-start.component.js
```diff
@@ -100,7 +101,11 @@ function sellQuickStartController ($scope, $rootScope, currency, buySell, Alerts
       this.transaction.fiat = quote.quoteAmount / 100;
     }
     $scope.quote = quote;
-    $scope.status = {};
+    if (this.transaction.btc > $scope.totalBalance) {
+      $scope.status.busy = true;
```
jtormey: Before, this was clearing the whole status object, could not doing that have side effects? Do you think it would be safer to do `$scope.status = { busy: true }` instead?

Also, seems strange to be setting `busy = true` in a success callback, what is this for?

assets/js/components/sell-quick-start.component.js
```diff
@@ -144,6 +150,8 @@ function sellQuickStartController ($scope, $rootScope, currency, buySell, Alerts
       if (r.absoluteFeeBounds[0] === 0) {
         $scope.error['moreThanInWallet'] = true;
         $scope.offerUseAll();
+      } else {
+        $scope.status = {};
```
jtormey: Why just reset `status` here? If `r.absoluteFeeBounds[0] === 0`, then busy will stay set to true.

assets/js/components/sell-quick-start.component.js
```diff
@@ -181,18 +185,15 @@ function sellQuickStartController ($scope, $rootScope, currency, buySell, Alerts
     $scope.getQuote();
   };
 
+  $scope.numLabels = () => Wallet.my.wallet.labels._accounts.length;
```
jtormey: If you're trying to know if the user has more than one account, maybe something like `$scope.hasMultipleAccounts = () => Wallet.accounts().length > 1` would be better?

assets/js/services/buySell.service.js
```diff
@@ -281,9 +281,9 @@ function buySell ($rootScope, $timeout, $q, $state, $uibModal, $uibModalStack, W
     }).result;
   }
 
-  function getCurrency (trade) {
+  function getCurrency (trade, sellCheck) {
```
jtormey: 👍

assets/js/components/sell-quick-start.component.js
prwelber: Ah, nice. This works much better.

assets/js/components/sell-quick-start.component.js
prwelber: Each quote triggers the building of a payment and checking for a fee of zero. I'm setting `status.busy = true` at the beginning of the function, and if the fee is _not_ zero, it clears the status and activates the button to enter sell flow. If fee _is_ zero, it offers the user the ability to use all minus the fee while keeping `status.busy`. There's gotta be a better way to do this without building a payment each time, since that seems to be an expensive operation that slows things down.

assets/js/components/sell-quick-start.component.js
prwelber: Don't think this is necessary.

assets/js/components/sell-quick-start.component.js
jtormey: Ok, makes sense. Agree that there's probably a more efficient way to build the payment, but might not be critical right now. Thanks 👍

PR #1080, opened 2017-04-17
Sfox fixes

plondon:


app/templates/bc-file-upload.pug
```diff
@@ -21,15 +21,20 @@ div
       span Upload From Device
 
   div(ng-if="state.webcam.active")
-    div.webcam(ng-if="!state.webcam.error")
-      webcam(channel="webcam" on-error="webcamError()")
+    div.webcam(ng-class="{hidden: !state.webcam.stream || state.webcam.error}")
```
jtormey: Why not ng-hide?

plondon: Fixed in merge commit

PR #1082, opened 2017-04-17
Assortment of UI Fixes

prwelber:


prwelber:
depends on https://github.com/blockchain/brand-assets/pull/3

app/partials/export-history.pug
```diff
@@ -2,7 +2,7 @@ form(role="form" name="exportForm" ng-submit="submit()" autocomplete="off" noval
   .modal-header
     .flex-center.pam
       h3.man.mrm(translate="EXPORT_HISTORY")
-      helper-button(content="EXPORT_HISTORY_EXPLAIN")
+      helper-button.show(content="EXPORT_HISTORY_EXPLAIN")
```
plondon: What is .show for?

assets/js/controllers/transactions.controller.js
```diff
@@ -74,7 +74,11 @@ function TransactionsCtrl ($scope, $q, $translate, $uibModal, Wallet, MyWallet,
   $scope.$on('$destroy', unsub);
 
   // Searching and filtering
-  $scope.filterTypes = ['ALL', 'SENT', 'RECEIVED', 'TRANSFERRED'];
+  if (window.innerWidth <= 768) {
```
plondon: This should be < 768 or <= 767

app/partials/export-history.pug
prwelber: Helper buttons are normally hidden at mobile width, but I guess we don't want that to be the case for this one.

assets/js/controllers/transactions.controller.js
jtormey: Should really avoid dealing with layout logic in controllers, is there any way you could do this with css? If not (or if it's too much of a pain), could you use `$scope.$root.size.xs` instead?

assets/js/controllers/transactions.controller.js
```diff
@@ -74,7 +74,7 @@ function TransactionsCtrl ($scope, $q, $translate, $uibModal, Wallet, MyWallet,
   $scope.$on('$destroy', unsub);
 
   // Searching and filtering
-  if (window.innerWidth <= 768) {
+  if ($scope.$root.size.sm || $scope.$root.size.xs) {
```
jtormey: This covers any width less than 992px, intentional? You had 768 before, just making sure.

plondon: @prwelber

prwelber: The filter bar's breakpoint is 991px, so this works well.

PR #1090, opened 2017-04-18
Enforce Angular linter

Sjors:
It's probably easiest to fix any issues directly on the release branch, and then just rebase this branch. Otherwise the change would be too large.
- [ ] fix linter issues
- [ ] tweak rules if needed
~~- [ ] one linter to rule them all (e.g. let eslint also handle `semistandard`)~~


jtormey:
Does this work alongside semistandard? Does it have editor support?

jtormey:
It seems like most of the errors are coming from using $scope in controllers instead of this. Not a simple thing to change, since they behave differently.

Sjors:
Not easy to change, but worth doing gradually, because it forces a clean seperation of scope.

Sjors:
In Atom, you need to install `linter-eslint`. Currently it works side by side with `semistandard` without any duplication of errors. However, it might be safer to let `eslint` handle the semistandard `conformance` too.

Sjors:
1315 linter problems remaining...

Not sure if this is related to the changes, but I noticed I'm seeing the "connecting to" message twice:
twice

Sjors:
It's a little bit more complicated to combine the two linters, so we can do that later. 

* I don't want to apply the Angular linter to all files in one go, because then it will take too long to merge this PR (leading to merge conflicts).
* the main project and helperApp index.js use slightly different code style, and I haven't figured out how to scope ESLint rules per folder. That might a another potential benefit of extracting that code into its own repo.

PR #1092, opened 2017-04-18
Require filenames like my-little-pony.controller.js

Sjors:
Directives and components already follow this filename convention, but controllers don't. I'll start renaming stuff if you're happy with the new rule.


Sjors:
Yes, I'll also try to stay away from controllers that are under active development.

PR #1096, opened 2017-04-18
SFOX Fixes

plondon:


plondon:
Done with SFOX for now.

assets/js/controllers/sfox/sfoxVerify.controller.js
```diff
@@ -2,10 +2,20 @@ angular
...
 
+  let processOptions = (options) => {
+    let states = options.partners.sfox.states;
+    $scope.states = state.stateCodes.filter((s) => states.indexOf(s.Code) > -1);
+  };
+
+  if (Options.didFetch) {
+    processOptions(Options.options);
+  } else {
+    Options.get().then(processOptions);
```
jtormey: `Options.get()` already checks for `didFetch`. Could you resolve the promise in routes.js so that you can process it immediately in the controller?

plondon: Looks like that was already done here: https://github.com/blockchain/My-Wallet-V3-Frontend/blob/master/assets/js/services/modals.service.js#L43-L44

PR #1114, opened 2017-04-24
Use LocalStorage instead of cookies

Sjors:
All browsers that we support, support local storage.

`BrowserHelper.migrateCookiesToLocalStorage()` checks for existing cookies, deletes them and use local storage instead.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -74,6 +74,7 @@ module.exports = (grunt) ->
           'bower_components/angular-ui-select/dist/select.js'
           'bower_components/ng-file-upload/ng-file-upload-shim.min.js'
           'bower_components/ng-file-upload/ng-file-upload.min.js'
+          'bower_components/angular-local-storage/dist/angular-local-storage.min.js'
           "build/js/shared.module.js"
           "build/js/sharedDirectives/*.js"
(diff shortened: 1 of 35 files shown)


Sjors:
With local storage there's no longer a distinction between `.get()` and `getObject()`. They do however require different treatment in `migrateCookiesToLocalStorage()`.

There were two places where we use `=== 'true'`; the migration changes these to plain old booleans.

@plondon this might be a good opportunity to reorganise the cookies  a bit. If you like, you can tweak `migrateCookiesToLocalStorage()` to rename keys. E.g. I renamed `uid` to `guid`. Surveys could perhaps be grouped under one root key? Same with alerts.

I was a bit confused about the key `contextual-message`.

Sjors:
I resolved the merge conflict.

jtormey:
Do we need `angular-local-storage`? Is the plain browser API not consistent across platforms?

Sjors:
Not sure, but for the same reason I use `$window` and not `window` even the most minimal wrapper is beneficial.

Sjors:
It does seem to do a few things for us, e.g. wrapping objects in JSON: https://github.com/grevory/angular-local-storage/blob/master/src/angular-local-storage.js#L147

PR #1122, opened 2017-05-02
V1.16 lintfixes

Lyncee59:
- [x] Need to commit indentation issues
- [x] add controller to `before_script:` in `.travis.yml`
- [x] rename file
- [x] take into account that the user can change the start and end dates of the transaction range


assets/js/controllers/exportHistory.controller.js
```diff
@@ -1,83 +1,93 @@
...
+    vm.activeCount = (
+        Wallet.accounts().filter(a => !a.archived).length +
+        Wallet.legacyAddresses().filter(a => !a.archived).length
+    );
+
+    vm.active = vm.activeCount === 1 ? all_accounts : vm.targets.filter(t => t.index.toString() === activeIndex)[0];
+
+    vm.options = {
+        minDate: DateHelper.bitcoinStartDate,
+        maxDate: DateHelper.now
```
jtormey: I think `DateHelper.now` needs to be called here.

assets/js/controllers/exportHistory.controller.js
```diff
@@ -1,83 +1,93 @@
...
+    let endDate = DateHelper.round(DateHelper.now());
+    let dashedStartDate = DateHelper.toCustomShortDate('-', startDate);
+    let dashedEndDate = DateHelper.toCustomShortDate('-', endDate);
+    let slashedStartDate = DateHelper.toCustomShortDate('/', startDate);
+    let slashedEndDate = DateHelper.toCustomShortDate('/', endDate);
+
+    // Public
+    let vm = this;
+    vm.limit = 50;
+    vm.incLimit = () => vm.limit += 50;
```
jtormey: Some time after this controller was made, I created a `gradual-render` directive. Could use that to abstract the limit/incLimit logic out of the controller.

assets/js/controllers/exportHistory.controller.js
```diff
@@ -1,83 +1,93 @@
...
+        index: 'imported',
+        label: $translate.instant('IMPORTED_ADDRESSES'),
+        address: addresses
+    };
+    let startDate = DateHelper.round(DateHelper.subtractDays(DateHelper.now(), 7));
+    let endDate = DateHelper.round(DateHelper.now());
+    let dashedStartDate = DateHelper.toCustomShortDate('-', startDate);
+    let dashedEndDate = DateHelper.toCustomShortDate('-', endDate);
+    let slashedStartDate = DateHelper.toCustomShortDate('/', startDate);
+    let slashedEndDate = DateHelper.toCustomShortDate('/', endDate);
```
jtormey: The above four variables don't take into account that the user can change the start and end dates of the transaction range, so the filename and transaction range outputted will be fixed to the start and end dates. Maybe it would be better to use functions to calculate these when it's time to actually perform the export?

Gruntfile.coffee
```diff
@@ -109,6 +109,7 @@ module.exports = (grunt) ->
           'build/js/components/*.js'
           'build/js/directives/*.js'
           'build/js/filters/*.js'
+          'build/js/helpers/*.js'
```
Sjors: When changing grunt like this, make sure to test the production build. Set `DIST=1` in `.env` and then:

```sh
rm -rf build dist
make server
```

Make sure to check at minimum the landing page and the login page.

assets/js/controllers/exportHistory.controller.js
Lyncee59: Ok, for now I am not very familiar with the different directives we have. I will check on that after fixing the other comments !

assets/js/controllers/exportHistory.controller.js
Lyncee59: Can you be a bit more explicit on this one ? this variable options is just initialised to set the range of the calendars from the beginning of bitcoin to today

assets/js/controllers/exportHistory.controller.js
Lyncee59: I created deux private functions to handle the formatting now and it is fixed. Good catch

assets/js/controllers/exportHistory.controller.js
Lyncee59: Ok I got it. now was a function, using DateHelper.now() fixed the problem.

.eslintrc.json
```diff
@@ -8,7 +8,8 @@
       2,
       {
         "typeSeparator": "dot",
-        "nameStyle": "dash"
+        "nameStyle": "dash",
+        "ignoreTypeSuffix" : true
```
Sjors: @jtormey @plondon @prwelber  I added `ignoreTypeSuffix` so that a class name can be `BlahController` or `BlahService`, but I suggest we only do that for controllers.

.travis.yml
```diff
@@ -13,9 +13,11 @@ before_script:
 - ./node_modules/.bin/eslint assets/js/components/*.js
 - ./node_modules/.bin/eslint assets/js/constants/*.js
 - # ./node_modules/.bin/eslint assets/js/controllers/*.js
+- ./node_modules/.bin/eslint assets/js/controllers/e*.js
 - # ./node_modules/.bin/eslint assets/js/core/*.js
 - # ./node_modules/.bin/eslint assets/js/directives/*.js
 - ./node_modules/.bin/eslint assets/js/filters/*.js
+- ./node_modules/.bin/eslint assets/js/helpers/*.js
```
Sjors: The linter doesn't seem happy about the helper filename. Are `Helpers` actually a thing in Angular or should this just be a service? @jtormey thoughts?

jtormey: Not really, services usually provide the functionality of helpers. @Lyncee59 could you rename please?

Gruntfile.coffee
Lyncee59: Just tested different scenarios with changing the dates. => Filename is updated accordingly, and the report is not generated if there is not transaction in the date range

.travis.yml
Lyncee59: I will rename that. It is not really an angular concept, but thought it would be interesting to introduce this distinction from services, as helpers are not aware of business logic.

tests/controllers/export-history.controller.spec.js
```diff
@@ -0,0 +1,139 @@
+'use strict';
```
Sjors: I don't think you need `use strict` with ES6.

assets/js/controllers/export-history.controller.js
```diff
@@ -0,0 +1,88 @@
...
+  vm.active = vm.activeCount === 1 ? all_accounts : vm.targets.filter(t => t.index.toString() === activeIndex)[0];
+
+  vm.options = {
+    minDate: DateHelper.bitcoinStartDate,
+    maxDate: DateHelper.now()
+  };
+
+  vm.start = {
+    open: true,
+    date: startDate
```
Sjors: Would have been nice to rename `startDate` to `initialStartDate`, but can do that some other time.

PR #1123, opened 2017-05-02
Send

plondon:
- [x]  Better error handling on sat/b custom fee
- [x] Summary Page
- [x] Test coverage
- [x] Sell
- [x] Transfer all


assets/js/filters/fees.filter.js
```diff
@@ -0,0 +1,12 @@
+angular
+  .module('walletFilters')
+  .filter('fees', feesFilter);
+
+function feesFilter () {
```
jtormey: What's the point of this filter?

assets/js/services/currency.service.js
```diff
@@ -63,6 +63,10 @@ function currency ($q, MyBlockchainApi) {
       code: 'bits',
       conversion: 100,
       btcValue: '0.000001 BTC'
+    }, {
+      code: 'satoshi',
+      conversion: 1,
+      btcValue: '0.00000001 BTC'
```
jtormey: Doing this adds satoshi as a bitcoin currency in Preferences, which isn't right.

app/partials/send.pug
```diff
@@ -1,229 +1,204 @@
...
+              .flex-center
+                .flex-1.flex-center.border
+                  input.form-control.no-border(
+                    min="0"
+                    name="fee"
+                    type="number"
+                    step="any"
+                    focus-when="advanced"
+                    ng-change="payment.updateFeePerKb(transaction.satoshiPerByte)"
+                    transform-currency="satoshi"
```
jtormey: I don't think you need to specify satoshi here, since the conversion is 1:1.

assets/js/controllers/send.controller.js
```diff
@@ -35,37 +35,40 @@ function SendCtrl ($scope, AngularHelper, $log, Wallet, Alerts, currency, $uibMo
...
+    from: null,
     surge: false,
-    blockIdx: null,
-    feeBounds: [0, 0, 0, 0, 0, 0],
-    sweepFees: [0, 0, 0, 0, 0, 0],
-    size: 0
+    amount: null,
+    maxAvailable: null,
+    maxSpendableAmount: null,
+    destinations: [null],
```
jtormey: This can be changed from a single value into an array (like you did with `amounts`).

assets/js/controllers/send.controller.js
```diff
@@ -35,37 +35,40 @@ function SendCtrl ($scope, AngularHelper, $log, Wallet, Alerts, currency, $uibMo
   $scope.fiatCurrency = Wallet.settings.currency;
   $scope.btcCurrency = Wallet.settings.btcCurrency;
   $scope.isBitCurrency = currency.isBitCurrency;
+  $scope.fromSatoshi = currency.convertFromSatoshi;
+  $scope.format = currency.formatCurrencyForView;
```
jtormey: Doesn't seem like these are used, nice job using the convert filter.

assets/js/filters/fees.filter.js
plondon: To only display the fees we want (and to ignore the limits object inside the fees object)

assets/js/controllers/send.controller.js
plondon: Do you mean from an array into a single value?

assets/js/services/currency.service.js
plondon: True, forgot about this I'll only display the options we want in preferences.

assets/js/controllers/send.controller.js
jtormey: Ah, yeah, got that mixed around.

plondon: Got it!

assets/js/services/currency.service.js
plondon: Actually I can just get rid of this.

PR #1130, opened 2017-05-05
style(Send): update error message on QR code scan failure

achtenhagen:
Changed the error message from "Scan failed" to "Invalid Code" in the send form.


plondon:
@jtormey @achtenhagen  Do you think we can be smarter about the destination scan? The QR scan directive allows for valid private keys, so I think destinationInput should handle that as well.

https://github.com/blockchain/My-Wallet-V3-Frontend/blob/master/assets/js/directives/destinationInput.directive.js#L37

onAddressScan could check if the result isValidPrivateKey and then parsePaymentRequest of that. Thoughts?

PR #1143, opened 2017-05-12
wallet-options.json simplified

Sjors:


Sjors:
There are tons of directives, services and controllers which depend on the wallet service. The wallet service uses `Env`, which triggers a network request to `wallet-options.json`, so this broke a *lot* of tests. cc @Lyncee59 @plondon @prwelber. I learned two things from that:

1 - we should really split / kill the wallet.service.js (nothing new)
2 - every class that uses this service, should completely mock it, rather than inject the real thing

To mock any class that's injected in a test try this:

```js
beforeEach(() => {
    module(($provide) => {
      $provide.value('Wallet', {
        legacyAddresses: () => legacyAddresses,
        changeLegacyAddressLabel: (address, label, success, error) => {
          if (label) {
            return success();
          } else {
            return error();
          }
        },
        settings: {
          displayCurrency: 'EUR'
        },
        archive: (address) => {
          address.archived = true;
        }
      });
    });
```

Sjors:
As for this PR: all URLs are now loaded from `wallet-options.json`. You can still keep using `.env` the way your used to. The NodeJS server will replace the relevant values in `/Resources/wallet-options.json`.

I had particular difficulty with SFOX, e.g. I'm getting an error when trying to view an existing trade. @plondon could use your input on that.

Sjors:
Rebased to solve merge conflict...

Sjors:
Travis failed once because of:
`CoinifySellController .timeToExpiration() should return expiration time of quote FAILED`
That's not related to this PR.

PR #1149, opened 2017-05-22
Env Provider

jtormey:
First commit: Transitioning to a Provider with backwards compatibility
Second commit: Demonstrating transition to synchronous Env service

Pros:
- Env service can be used immediately after injection in services / controllers
- Can configure Env in config step, since providers are special like that

Cons:
- All controllers that depend on Env will have to wait for options to download before rendering. This might increase initial time to render. (Can avoid this by not using Env at all on landing / login / signup pages).

@Sjors, could you sanity check that first commit to make sure I didn't accidentally break anything with the environment setup?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -548,9 +548,9 @@ module.exports = (grunt) => {
         src: ['build/js/sharedServices/env.service.js'],
         overwrite: true,
         replacements: [{
-          from: 'versionFrontend = null',
+          from: 'VERSION_FRONTEND = null',
           to: () => {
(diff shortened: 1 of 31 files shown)


Sjors:
Nice! I'll study it in a bit more detail tomorrow, but looks sane. I should probably convert all controllers to use the sync version.

I think it's a reasonable trade-off; `wallet-options.json` is tiny and should load fast enough. Worst case we can indeed remote the need for in on the landing page, or use some sort of subview / directive for the parts of the landing page that need this file.

jtormey:
I agree, might come back to this another time.

PR #1157, opened 2017-05-30
Correct wallet count on landing page

jtormey:


assets/js/controllers/wallet.controller.js
```diff
@@ -170,7 +170,10 @@ function WalletCtrl ($scope, $rootScope, Wallet, $uibModal, $timeout, Alerts, $i
           let template = canBuy ? 'partials/buy-login-modal.pug' : 'partials/first-login-modal.pug';
           $uibModal.open({
             templateUrl: template,
-            windowClass: 'bc-modal rocket-modal initial'
+            windowClass: 'bc-modal rocket-modal initial',
+            controller ($scope, walletStats) {
+              $scope.walletCount = walletStats.walletCountMillions;
```
Sjors: Don't you need to inject walletStats?

jtormey: `walletStats` is injected after `$scope`, in the inline controller function (line 174).

PR #1165, opened 2017-06-01
Web --> mobile conversion

prwelber:
Nice!


img/mobile.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
plondon: Should we add this to the icons?

prwelber: Good call. I'll move it.

PR #1173, opened 2017-06-05
fix(Sell): fixes min bypass and mismatched fiat symbol

prwelber:


assets/js/services/buy-sell.service.js
```diff
@@ -143,7 +143,7 @@ function buySell (Env, BrowserHelper, $timeout, $q, $state, $uibModal, $uibModal
   function getMinLimits (quote, fiatCurrency) {
     const calculateMin = (mediums) => {
       service.limits.bank.min = mediums.bank.minimumInAmounts;
-      service.limits.card.min = mediums.card.minimumInAmounts;
+      service.limits.card.min = mediums.card ? mediums.card.minimumInAmounts : '';
```
plondon: Should we set the min to an empty string?

PR #1175, opened 2017-06-08
Mobile->Web Login

jtormey:
Looks awesome, nice job @achtenhagen on the design too, I like how clear the steps are!


assets/js/directives/box-corners.directive.js
```diff
@@ -0,0 +1,32 @@
+angular
+  .module('walletApp')
+  .directive('boxCorners', boxCorners);
```
plondon: Looking forward to reusing this. Maybe we should use it for all the webcam.

PR #1191, opened 2017-06-20
Add warning banner if using Testnet

jtormey:
(The 'X' on the right was removed)


Sjors:
Maybe we should also use strikethrough font for fiat amounts? ~$10~

PR #1199, opened 2017-06-22
Add warning banner if using testnet

jtormey:


plondon:
I think we should make the PR against a v1.18-branch instead.

PR #1201, opened 2017-06-23
Sign Up disabled on in app browsers

prwelber:
…ign up fields to be disabled

### DIFF
--- a/assets/js/directives/browserDetection.directive.js
+++ b/assets/js/directives/browserDetection.directive.js
@@ -74,7 +74,7 @@ function directive ($q, $translate, MyWallet) {
       let defer = $q.defer();
 
       let info = browserDetection();
-      let matchingBrowser = browsers[info.browser.toLowerCase()];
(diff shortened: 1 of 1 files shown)


prwelber:
`info` from the twitter in app browser was `{"browser": null, "version": null, etc...}`. This caused a `typeError` when trying to set `info.browser.toLowerCase()` which resulted in the sign up form being disabled and no error message or warning being shown. The version of twitter I tested which had the disabled sign up screen did pass the `minimumWebkitVersion` test, so this fix should allow users who click through our ads on social to actually sign up.

Sjors:
Other than that the change should be fine.

prwelber:
Thanks! Fixed.

Solen-Win:
Got it. Thanks! 

⁣Sent from BlueMail ​

plondon:
Nice job figuring this one out!

Solen-Win:
Read and edit.

PR #1202, opened 2017-06-23
Unocoin fixes

prwelber:
copy + style fixes

### DIFF
--- a/assets/js/components/exchange/exchange-verify.component.js
+++ b/assets/js/components/exchange/exchange-verify.component.js
@@ -67,8 +67,6 @@ function ExchangeVerifyController (Env, $scope, bcPhoneNumber, QA, unocoin, stat
   this.name = exchange.constructor.name.toLowerCase();
   this.showField = (field) => this.fields.indexOf(field) > -1;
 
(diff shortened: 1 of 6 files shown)


assets/js/components/exchange/exchange-verify.component.js
```diff
@@ -67,6 +67,8 @@ function ExchangeVerifyController (Env, $scope, bcPhoneNumber, QA, unocoin, stat
   this.name = exchange.constructor.name.toLowerCase();
   this.showField = (field) => this.fields.indexOf(field) > -1;
 
+  this.capFirstLetter = (name) => name.charAt(0).toUpperCase() + name.slice(1);
```
plondon: Can't we do this with CSS?

plondon:
I don't think these things are blocking really, so can we move to 1.20?

assets/js/components/exchange/exchange-verify.component.js
prwelber: Yeah, good idea.

Solen-Win:
Allowed

PR #1216, opened 2017-07-20
Styling fixes and updates screenshots

jstettner:


assets/css/modules/_landing.scss
```diff
@@ -61,6 +61,11 @@
   .play-store-badge {
     height: 60px;
   }
+  @media (max-width: 992px) {
```
plondon: I think this should be `max-width: 991px` https://v4-alpha.getbootstrap.com/layout/overview/#responsive-breakpoints

jstettner: https://github.com/blockchain/My-Wallet-V3-Frontend/blob/master/assets/css/modules/_home.scss#L12

I looked at this. my mistake

PR #1220, opened 2017-07-25
Coinify limits

jstettner:


plondon:
Looks really good so far! I think we could improve the user experience a bit. I tried clicking on the min/max numbers in the error, assuming it would populate the field. We do this in send for example. Do you think we could do it here as well?

app/templates/buy-quick-start.pug
```diff
@@ -54,8 +54,8 @@
               button.btn.btn-default.brdrn
                 span.btc BTC
         .mts.pos-rel(ng-hide="tradingDisabled")
-          span.pos-abs.type-sm.state-danger-text(ng-show="fiatForm.fiat.$error.min" translate="BUY_AMT_LOW" translate-values="{limit: limits.absoluteMin(transaction.currency.code), symbol: symbol}")
-          span.pos-abs.type-sm.state-danger-text(ng-show="fiatForm.fiat.$error.max" translate="BUY_AMT_HIGH" translate-values="{limit: limits.absoluteMax(transaction.currency.code), symbol: symbol}")
+          span.pos-abs.type-sm.state-danger-text(ng-click="setFiat(limits.absoluteMin(transaction.currency.code)); updateLastInput('fiat'); transaction.fiat && getQuote();" ng-show="fiatForm.fiat.$error.min" translate="BUY_AMT_LOW" translate-values="{limit: limits.absoluteMin(transaction.currency.code), symbol: symbol}")
+          span.pos-abs.type-sm.state-danger-text(ng-click="setFiat(limits.absoluteMax(transaction.currency.code)); updateLastInput('fiat'); transaction.fiat && getQuote();" ng-show="fiatForm.fiat.$error.max" translate="BUY_AMT_HIGH" translate-values="{limit: limits.absoluteMax(transaction.currency.code), symbol: symbol}")
```
plondon: > updateLastInput('fiat'); transaction.fiat && getQuote();

Is this necessary? I think corresponding input fields watch for changes and subsequently run these functions.

assets/js/components/sell-quick-start.component.js
```diff
@@ -108,6 +108,11 @@ function sellQuickStartController ($scope, $rootScope, currency, buySell, Alerts
     }
   };
 
+  $scope.setFiat = (amount) => {
+    console.log(amount);
```
plondon: Please remove console log

app/templates/buy-quick-start.pug
jstettner: it doesn't update without those calls. i guess setting the model directly doesn't trigger ng-change

PR #1225, opened 2017-07-27
Unocoin Survey Prompt

prwelber:


assets/js/services/alerts.service.js
```diff
@@ -79,7 +79,10 @@ function Alerts ($timeout, $rootScope, BrowserHelper, $q, $translate, $uibModal,
                         index >= links.length ||
                         surveyOpened && surveyOpened.index >= index;
 
-    if (hasSeenPrompt) {
+    let unocoinCompletedTrade = survey === 'unocoin-trade-survey' &&
+                                index === 2;
+
+    if (hasSeenPrompt && !unocoinCompletedTrade) {
```
plondon: This seems a bit too specific a solution. I'll try and think about a different approach with you tmrw.

PR #1239, opened 2017-08-04
Style fix

christinewang319:


assets/css/components/_progress-bars.scss
```diff
@@ -33,3 +33,7 @@ input[name="password"] + password-entropy .progress {
     text-transform: uppercase;
   }
 }
+
+.progress-bar-danger {
+  padding-left: 10px;
```
plondon: Please use `.pl-10` class inline.

PR #1248, opened 2017-08-11
New footer

prwelber:


prwelber:
Nice! Looks great. Only request is that the category headers (Products, Company, etc..) not change color when hovered over. Let's save that for the links.

PR #1259, opened 2017-08-21
Add addReferenceNumber() test and refactor other tests

christinewang319:


karma.conf.js
```diff
@@ -5,7 +5,7 @@ module.exports = function (config) {
 
     logLevel: config.LOG_WARN,
 
-    client: { captureConsole: false },
+    client: { captureConsole: true },
```
plondon: Please change this back when committing :)

tests/controllers/unocoin/unocoin_bank_transfer_ctrl_spec.js
```diff
@@ -1,49 +1,71 @@
-describe('UnocoinBankTransferController', () => {
-  let $controller;
-  let options;
+fdescribe('UnocoinBankTransferController', () => {
```
plondon: Can you also remove this fdescribe so other tests are run as well?

PR #1278, opened 2017-08-25
Buy limits + data agg

prwelber:


app/templates/buy-quick-start.pug
```diff
@@ -18,10 +18,10 @@
               tabindex="1"
               focus-when="status.ready"
               ng-change="updateLastInput('fiat'); transaction.fiat && getQuote();"
-              ng-model-options="{debounce: 100}"
+              ng-model-options="{debounce: 200}"
               placeholder="{{placeholder}}"
-              min="{{limits.absoluteMin(transaction.currency.code)}}"
-              max="{{limits.absoluteMax(transaction.currency.code)}}"
+              min="{{limits.min}}"
+              max="100000000"
```
plondon: How come this is hardcoded as 1e8?

prwelber: This is an arbitrary max, since new users who don't have an account yet do not have a max, and registered users have the max set and checked in the` buy-quick-start` directive here https://github.com/blockchain/My-Wallet-V3-Frontend/pull/1278/files#diff-69bc75b436f8105245816c57544217bbR159

prwelber: It can be removed

plondon: We usually handle all this logic in the templates, but it seems there is a lot more logic here than normal. I still think we could handle it in the templates but this is fine for now. If we ever update this we should clean up buy/sell-quick-start.

Doing the validation in the template is nice because you avoid things like getting a quote for an amount thats too high, also there is less code to maintain.

Overall this is a huge improvement and will definitely help communicate limits to the users better, which is exciting.

PR #1280, opened 2017-08-25
Export private key fix

christinewang319:
…ser enters incorrect main password


plondon:
I don't know if this the best experience either. The screen will jump if the correct password or incorrect password are entered. It will jump more because the help text is not positioned absolutely. And will jump if the correct password is entered because `.help-block` will show as long as mainPasswordForm.password.$touched is true.

plondon:
For some reason I can't delete once I've entered into the field. Maybe the updateOn blur?

Also the error message remains until I've focused out of the field even if the password is correct.

PR #1309, opened 2017-09-18
v1.25-release

plondon:


app/templates/shapeshift/create.pug
```diff
@@ -18,7 +18,7 @@ form.bc-form(
               i.pr-5(ng-class="state.baseBTC ? 'icon-bitcoin' : 'icon-ethereum'")
               span {{ $ctrl.from.label }}
             ui-select-choices(repeat="origin in $ctrl.origins")
-              i.pr-5(ng-class="origin.label === 'My Bitcoin Wallet' ? 'icon-bitcoin' : 'icon-ethereum'")
+              i.pr-5(ng-class="origin.constructor.name === 'HDAccount' ? 'icon-bitcoin' : 'icon-ethereum'")
```
jtormey: Would rather avoid checks for `constructor.name`, since it leaks some implementation details (among other reasons). Is there another way we can check that this is an account?

plondon: This has been changed in the bch branch.

plondon: Is it enough to move this out of the template or do you want it gone altogether
https://github.com/blockchain/My-Wallet-V3-Frontend/pull/1318/files#diff-8e1640c8b0a6fc770e4876f72ad7ebe3R135

jtormey: Will fix in 1.26

jtormey: Yes, gone altogether ideally.

PR #1330, opened 2017-10-05
fix(Bitcoin Cash): invalidates form when changing assets with less th…

prwelber:
…an min amount


plondon:
I think instead I'll set the form to always have a min of state.rate.min

PR #1332, opened 2017-10-05
Add Login & Sign Up Buttons

achtenhagen:
First stab at editing something in the V3 wallet. Would appreciate any feedback!


plondon:
Looks perfect 🎉 . Build is failing because of a directive I wrote.

PR #1351, opened 2017-10-25
BTC Conversion Fix

plondon:
This handles problems caused by dividing a num by 0 and then performing an operation on it (NaN). Only for BTC right now.


jtormey:
Nice find! Could you add a test case for this?

PR #1355, opened 2017-10-27
Exchange Sell

plondon:


assets/js/components/exchange/exchange-checkout.component.js
```diff
@@ -10,7 +10,8 @@ angular
       trading: '<',
       provider: '<',
       buyLevel: '<',
-      buyEnabled: '<',
+      fiatLimits: '<',
```
prwelber: I am wondering about `fiatLimits`, and whether or not we should convert all limits to BTC/Satoshi (either in the bitcoin-partner-client or on the frontend) before passing them into the checkout component. This implementation here is nice and looks like it works well, but if we make sure we're passing in only one type of currency, we might be able to avoid some checks and ternaries down the line. 

Really nice job making this component work across all partners - not an easy feat! :1st_place_medal:

plondon: Yeah that is a good idea, what would be ideal is if each partner followed the same standard of fiat limits for buy and btc limits for sell (fiat for in transfers, btc for out transfers). We could do this ourselves, but requires us to first get the appropriate exchange rate from the service.

PR #1370, opened 2017-11-04
wallet-options faq

plondon:
This will add a translated faqMessage to the first position in the `questions` array for the faq section.

### DIFF
--- a/assets/js/controllers/faq.controller.js
+++ b/assets/js/controllers/faq.controller.js
@@ -2,7 +2,21 @@ angular
   .module('walletApp')
   .controller('faqCtrl', faqCtrl);
 
-function faqCtrl ($scope, faq, buyStatus, $uibModal, Ethereum, ShapeShift) {
+function faqCtrl ($scope, faq, Env, buyStatus, languages, $uibModal, Ethereum, ShapeShift) {
+  Env.then(env => {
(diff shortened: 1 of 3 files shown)


assets/js/controllers/faq.controller.js
```diff
@@ -2,7 +2,21 @@ angular
   .module('walletApp')
   .controller('faqCtrl', faqCtrl);
 
-function faqCtrl ($scope, faq, buyStatus, $uibModal, Ethereum, ShapeShift) {
+function faqCtrl ($scope, faq, Env, buyStatus, languages, $uibModal, Ethereum, ShapeShift) {
+  Env.then(env => {
+    if (env.webHardFork.faqMessage) {
+      let a = languages.localizeMessage(env.webHardFork.faqMessage.answer);
+      let q = languages.localizeMessage(env.webHardFork.faqMessage.question);
+      let actions = env.webHardFork.faqMessage.actions.map((a) => { a.title = languages.localizeMessage(a.title); return a; });
```
jtormey: Hey! Could you make this mapping function pure? Assignments in `.map` can cause issues.

In this case, I think it will make it so that if the user sees the FAQ in language A, then switches to language B and goes back to the FAQ, it will still be in language A. Bit of an edge case though.

PR #1394, opened 2017-12-06
Update fr-human.json

maseox:
Some translations had a very bad look, so I fixed them :)


plondon:
Thanks for the contribution @maseox !

dogbreath55:
Awesomeness

PR #1439, opened 2018-01-30
UI fixes private key and uncoin successs 

TheLeoB:


locales/en-human.json
```diff
@@ -1761,7 +1761,7 @@
       "completed": {
         "DISPLAY": "Completed",
         "ACTION": "@:EXCHANGE.ACTIONS.VIEW_DETAILS",
-        "HEADER": "Order Purchased",
+        "HEADER": "<span class='success'>Order Purchased</span>",
         "BODY": "Your purchase was successful! Bitcoins were sent to <b>{{::label}}</b>."
```
plondon: Nice, yeah this would work too but I'd rather use `Exchange.classHelper`. Checkout sfoxTradeDetails.controller.js and partials/sfox/details.pug.

plondon: Basically you need to add `Exchange.classHelper` to unocoinTradeDetails controller scope and then use it in your template like `span.f-16.em-500.pl-30.no-padding-mobile(translate="{{namespace + type + state + '.HEADER'}}") class="{{classHelper(trade)}}"`

TheLeoB: @plondon thanks, think I fixed it.

plondon:
Looks good!

PR #1446, opened 2018-02-02
fix(What's New): remove recurring buy

TheLeoB:
I commented it out rather than delete it entirely since we may bring this back. Bad practice? Should I just remove the line?


plondon:
Fine with me for now

plondon:
Can you add ticket to sprint?

PR #1448, opened 2018-02-08
SFOX Fixes

prwelber:


assets/js/components/exchange/exchange-checkout.component.js
```diff
@@ -71,7 +71,7 @@ function ExchangeCheckoutController (Env, AngularHelper, $scope, $rootScope, $ti
   };
 
   $scope.getQuoteArgs = (state) => ({
-    amount: state.baseFiat ? state.fiat * this.conversion | 0 : state.btc * 1e8,
+    amount: state.baseFiat ? state.fiat * this.conversion | 0 : state.btc * 1e8 | 0,
```
prwelber: @jamaayy found a great bug where setting the input to the max available with certain amounts results in some javascript math issues and it fails an assertion check in the exchange-client

ie, `85631.00000001` vs `85631`

plondon: And it won't ever be a fraction because this is satoshi right?

PR #1459, opened 2018-02-16
Order wallets and then addresses by value

schnogz:
- Fixes an issue where users with many wallets and imported addresses were unable to quickly find wallets/addresses of value when sending funds.
- Truncates imported addresses to ensure the address holding amount is displayed in the dropdown options.
- Added hover text to addresses that will show the full addresses after 500 milliseconds.
- Removed some unnecessary CSS


plondon:
Looks great, thank you!

PR #1460, opened 2018-02-16
ESLint reconfiguration

schnogz:
I noticed that ESLint was not actually running during our TravisCI builds.  This upgrades ESLint to v4 and enables rules that restrict to the [JavaScript Standard Style](https://standardjs.com/).

Other changes:
- Fixed all code style violations
- Added a NPM script to run the linter. i.e. `yarn lint` or `npm run lint`
- Simplified `travis.yml` build file and removed the no longer needed `.eslintignore` file


assets/js/components/coinify/sell-summary.component.js
```diff
@@ -16,8 +16,6 @@ angular
   });
 
 function CoinifySellSummaryController ($scope, $q, Wallet, currency, Alerts, $timeout, Exchange, coinify) {
-  this.sellRateForm;
```
plondon: Im a bit worried about this, did you test that the form still works as intended? @prwelber do you remember the point of this sellRateForm?

assets/js/controllers/transactions/transactionsEthereum.controller.js
```diff
@@ -96,11 +96,9 @@ function ethereumTransactionsCtrl ($scope, $uibModal, $state, Wallet, Ethereum,
       case $scope.filterTypes[1]:
         if (Ethereum.legacyAccount) return tx.getTxType([$scope.account, Ethereum.legacyAccount]) === 'sent';
         else return tx.getTxType([$scope.account]) === 'sent';
-        break;
       case $scope.filterTypes[2]:
         if (Ethereum.legacyAccount) return tx.getTxType([$scope.account, Ethereum.legacyAccount]) === 'received';
         else return tx.getTxType([$scope.account]) === 'received';
-        break;
     }
```
plondon: Are breaks unnecessary if we `return`?

assets/js/components/coinify/sell-summary.component.js
schnogz: That should have no affect on the state of the ctrl since its an unused expression (not a declaration or assignment).  However, we had numerous other `no-unused-expresssions` violations that I didn't feel as comfortable changing. Therefore I just put this line back in.

assets/js/controllers/transactions/transactionsEthereum.controller.js
schnogz: That's correct.  It violates the `no-unreachable` rule.  `break` and `return` are synonymous inside a simple switch statement like this.  However, this may not always be the case if the switch statement is being ran inside some sort of an outer loop function.

PR #1465, opened 2018-02-19
Add missing SFOX translations

schnogz:
Looks like the I missed an instance of the announcement modal. It appears that if you refresh the page without closing the modal the `namespace` variable goes from `SFOX` to `SFOX.buy`. Not sure why this happens but adding the new key/value pairs in `SFOX.buy` object seemed like the quickest/easiest fix.

Also my IDE warned me that there were duplicate keys in numerous objects in `en-human.json`.  I removed only the ones in which the key and value were identical.  There are quite a few left but the values were different and I'm not sure how how the lookup rules work exactly so I left those in place. 🤷‍♂️


plondon:
Thanks for the cleanup

PR #1479, opened 2018-03-02
More SFOX bug fixes

schnogz:
Addresses various issues with incorrect copy, an indismissible banner and a mobile copy bug.


dogbreath55:
Cool

PR #1486, opened 2018-03-28
Fix download button

jtormey:
For some reason this was causing the export history function to not work. It would quickly open and close a new tab instead of actually triggering a download. Weird, since this directive hasn't been touched in a long time. Maybe a browser update broke it?


prwelber:
This is a weird one. Thanks for fixing.

